PR #16, opened 2017-07-06
Add build instructions to readme

ruuda:
Not everybody who might be interested in using this will be familiar with the Haskell build tools, so write a bit about that.


duijf:
Thanks @ruuda! :)

PR #17, opened 2017-07-06
Mention `stack exec`

drewr:
A nice shortcut to running the built binary.

Great project btw! Very similar to a GPG-based tool we use internally that we hadn't expanded to out Vault usage yet. This works very well.


duijf:
Hi Drew! Thank you very much for the PR and the kind words :smile:

wesleybowman:
@drewr strange to see a fellow Tennesseean in this repository :+1:

PR #19, opened 2017-08-25
Fail if duplicate environment variables are found

### LINKED ISSUE #5: Error/print warning when secret variable occurs more than once
Since we now support custom variables (#3), users can make the mistake of assigning to a variable more than once. We need to either error out or print a warning in this case

futtetennista:
Fixes #5 

In the issue the options mentioned are failing or printing a warning: I leaned towards the former since overriding a variable name is most of the times something done by accident in my experience. But let me know what you folks think about it, if you have a different opinion it won't take much work to change the PR.

The time complexity of `dups` is still quadratic but since that wasn't seen as a performance issue before I didn't use a better one.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -157,7 +157,7 @@ main = do
       dups (x:xs) | isDup x xs = Left x
                   | otherwise = dups xs
 
-      isDup x = foldr (\y acc -> x == y || acc) False
+      isDup x = foldr (\y acc -> acc || x == y) False
 
 
 parseSecret :: String -> Either String Secret
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -54,6 +54,7 @@ data VaultError
   | ServerUnavailable LBS.ByteString
   | ServerUnreachable
   | InvalidUrl        Secret
+  | DuplicateVar      String
   | Unspecified       Int LBS.ByteString
 
 --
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -130,17 +130,29 @@ main = do
   newEnvOrErrors <- Async.mapConcurrently (requestSecret opts) secrets
 
   let
+    checkNoDuplicates e =
+      let keys = map fst (e ++ env)
+      in case dups keys of
+           Left x -> errorWithoutStackTrace $ "Found duplicate variable \"" ++ x ++ "\""
```
arianvp: Justa driveby comment. @duijf perhaps we cant to switchj away from `errorWithoutStackTrace` and move to `ExceptT/Either` ?  Partial functions are never fun.

duijf: We can look into that! I do prefer merging this PR first though

duijf: Right now, we do use `Either` and only run `errorWithoutStackTrace` in `main`. That won't change if we put this logic in `Either`/`ExceptT`, right? (Can you open an issue where you make the case for this if you think this is important?)

app/Main.hs
```diff
@@ -130,17 +130,29 @@ main = do
...
+           _ -> e ++ env
     newEnv = case sequence newEnvOrErrors of
       -- We need to eliminate duplicates in the environment and keep
```
duijf: Can you update this comment? `nubBy` has been removed

futtetennista:
Thanks for the feedback! I need to solve the conflicts with master before this can be merged, I could either merge this branch or rebase it and then force push it to have a cleaner history: do you folks have any preferences?

duijf:
Thanks for the quick improvements, I'll take a look whenever you think they're ready! 

We prefer rebase + force push :smile:

futtetennista:
@duijf I believe it's now ready for a review 🙂

app/Main.hs
arianvp: The case is that partial functions are always evil :tm: but i'll make a separate issue for it :)

futtetennista:
I agree. :) I read your comments above but since that wasn't the goal of
this PR I preferred to keep it simple and focused.

duijf:
Merged! Thanks a lot for addressing the comments and adding the integration test! :heart:

futtetennista:
Glad I was able to contribute…and have fun in the process ;-)

PR #22, opened 2017-09-05
[Prototype] Avoid impure exceptions

futtetennista:
In #19 @arianvp [mentioned](https://github.com/channable/vaultenv/pull/19#discussion_r135458298) that it would be desirable to avoid `errorWithoutStackTrace` in favour of `ExceptT` and @duijf and myself seemed to agree on it.  I managed to have a first working solution that implements that change: could you folks give me some feedback? Does this go in the direction you envisioned or would you implement it some other way?

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -192,7 +192,7 @@ parseSecret line =
 
 readSecretList :: (MonadError VaultError m, MonadIO m) => FilePath -> m [Secret]
 readSecretList fname = do
-  esecrets <- liftIO $ fmap (sequence . fmap parseSecret . lines) (readFile fname)
+  esecrets <- liftIO $ traverse parseSecret . lines <$> readFile fname
   either (throwError . IOError) return esecrets
 
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,7 +1,8 @@
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 import Control.Lens           (preview)
-import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Char
 import Data.Either            (isLeft)
 import Data.List              (findIndex)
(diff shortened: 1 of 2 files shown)


rightfold:
What does "[Spike]" mean?

app/Main.hs
```diff
@@ -214,6 +226,12 @@ requestSecret opts secret =
     Retry.retrying vaultRetryPolicy shouldRetry retryAction
 
 
+requestSecrets :: (MonadError VaultError m, Traversable t, MonadIO m)
+               => Options -> t Secret -> m (t EnvVar)
+requestSecrets opts secrets = do
+  esecretEnv <- liftIO $ Async.mapConcurrently (requestSecret opts) secrets
+  either throwError (return . id) $ sequence esecretEnv
```
rightfold: Why compose with `id`?

app/Main.hs
```diff
@@ -180,8 +190,10 @@ parseSecret line =
                 }
 
 
-readSecretList :: FilePath -> IO (Either String [Secret])
-readSecretList fname = fmap (sequence . fmap parseSecret . lines) (readFile fname)
+readSecretList :: (MonadError VaultError m, MonadIO m) => FilePath -> m [Secret]
+readSecretList fname = do
+  esecrets <- liftIO $ fmap (sequence . fmap parseSecret . lines) (readFile fname)
+  either (throwError . IOError) (return . id) esecrets
```
rightfold: This seems to be a parse error, not an I/O error.

app/Main.hs
```diff
@@ -180,8 +190,10 @@ parseSecret line =
                 }
 
 
-readSecretList :: FilePath -> IO (Either String [Secret])
-readSecretList fname = fmap (sequence . fmap parseSecret . lines) (readFile fname)
+readSecretList :: (MonadError VaultError m, MonadIO m) => FilePath -> m [Secret]
+readSecretList fname = do
+  esecrets <- liftIO $ fmap (sequence . fmap parseSecret . lines) (readFile fname)
```
rightfold: `sequence . fmap f` = `traverse f`

app/Main.hs
rightfold: Why compose with `id`?

app/Main.hs
futtetennista: Thanks that's useless.

futtetennista:
@rightfold I was under the assumption that it was a synonym of "prototype", we used it a lot in the company I used to work for but I just double checked the English dictionary and there's nothing even remotely close to that meaning 😳 weird, my apologies.

app/Main.hs
futtetennista: yes you're right. I got confused because I was also thinking: what happens if `readFile` fails? I've taken into account that possibility in one of my next commits.

rightfold:
Thank you for your contribution, which I have just merged. GitHub did not notice, so I'm closing this PR.

Issue #23, opened 2017-09-18
Vaultenv should not rewrite error messages


NunoAlexandre:
We currently capture _all_ http exceptions as a `ServerUnreachable` with no additional info. This is not useful at all. If the token is invalid, or the secret does not exist, there is no way to debug information.

We should not be clever and try to print 'pretty' errors. Just print the raw exception so we can figure out what is wrong.

https://github.com/channable/vaultenv/blob/7d16472a54c944c8c55140302a21dc4f9649b53b/app/Main.hs#L225

arianvp:
SORRY FOR THE ALL CAPS BUT THERE IS CURRENTLY A BUG IN GITHUB THAT BREAKS TYPING.

WE SHOULD ALSO MAKE EXTOERR TOTAL. CURRENTLY IF THERE IS AN UNEXPECTED EXCEPTION. EXTOERROR CRASHES WITH A IREFUTABLE PATTERN ERROR. THIS IS NOT GOOD. LETS MAKE A CATCHALL NAMED "UNCAUGHTEXCEPTION" THAT PATTERN MATCHES ON CONTROL.EXCEPTION.SOMEEXCEPTION

NunoAlexandre:
> WE SHOULD ALSO MAKE EXTOERR TOTAL. CURRENTLY IF THERE IS AN UNEXPECTED EXCEPTION. EXTOERROR CRASHES WITH A IREFUTABLE PATTERN ERROR. THIS IS NOT GOOD. LETS MAKE A CATCHALL NAMED "UNCAUGHTEXCEPTION" THAT PATTERN MATCHES ON CONTROL.EXCEPTION.SOMEEXCEPTION

The  `exToErr` function signature is `:: HttpException -> VaultError`.  `HttpException` only has two constructors, the two that are now pattern-matched, so the function IS total. 

I can still add that coverage for future purposes (if a new constructor gets added to the `HttException` type) but I don't find that necessary.

PR #24, opened 2017-09-18
Print the raw exception on 'ServerUnreachable'

NunoAlexandre:
### Addressing https://github.com/channable/vaultenv/issues/23 (Vaultenv should not rewrite error messages)

The change simply involved adding a field of type `HttpException` to `ServerUnreachable` error and adjust affected code PLUS rewriting the error message, now not guessing but printing the raw exception we get from vault. 

> WE SHOULD ALSO MAKE EXTOERR TOTAL. CURRENTLY IF THERE IS AN UNEXPECTED EXCEPTION. EXTOERROR CRASHES WITH A IREFUTABLE PATTERN ERROR. THIS IS NOT GOOD. LETS MAKE A CATCHALL NAMED "UNCAUGHTEXCEPTION" THAT PATTERN MATCHES ON CONTROL.EXCEPTION.SOMEEXCEPTION

The  `exToErr` function signature is `:: HttpException -> VaultError`.  `HttpException` only has two constructors, the two that are now pattern-matched, so the function IS total. 

I can still add that coverage for future purposes (if a new constructor gets added to the `HttException` type) but I don't find that necessary.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -52,7 +52,7 @@ data VaultError
   | Forbidden
   | ServerError       LBS.ByteString
   | ServerUnavailable LBS.ByteString
-  | ServerUnreachable
+  | ServerUnreachable HttpException
   | InvalidUrl        Secret
   | DuplicateVar      String
   | Unspecified       Int LBS.ByteString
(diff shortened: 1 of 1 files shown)


arianvp:
lgtm. I can type without caps lock again as well :)

NunoAlexandre:
Thanks

> I can type without caps lock again as well :)
Awesome! :P

Issue #13, opened 2017-06-30
allow to set token using VAULT_TOKEN env variable


arianvp:
This would make `vaultenv` play nice with https://nomadproject.io/  as your nomad cluster will automagically inject the right token for vaultenv https://www.nomadproject.io/docs/job-specification/vault.html

PR #26, opened 2017-09-19
Use VAULT_TOKEN environment variable if --token is not supplied

fatho:
Addresses issue #13.


app/Main.hs
```diff
@@ -101,13 +102,15 @@ optionsParser = Options
        <*> switch
            (  long "no-inherit-env"
            <> help "don't merge the parent environment with the secrets file")
+  where
+    environ vars key = maybe mempty value (lookup key vars)
```
rightfold: `Data.Foldable.foldMap = maybe mempty`, not Very important though.

Issue #18, opened 2017-08-15
Cache DNS Lookup 


arianvp:
Some of our components require hundreds of secrets to be collected from vault at once.   (We have hundreds of services requesting hundreds of secrets in bursts sometimes)  This was actually hitting limits of our DNS resolver.   Simply doing the lookup once, and then use it for consequent http requests would solve this.

arianvp:
I think we should solve this at the OS level.    There is probably a caching parameter that we're missing.

However,  I think we could reduce the amount of requests that we send for a secrets file by only requesting each path once.    I'll file a new issue for that.

rkrzr:
> However, I think we could reduce the amount of requests that we send for a secrets file by only requesting each path once.

Do you mean for each "store" (like e.g. `secret/sentry`)? That's a good idea.

arianvp:
@rkrzr yes!  exactly.

duijf:
In our production setup, we currently cache DNS lookups on the host level. This is IMO the better place for this: vaultenv shouldn't attempt to cache this. Otherwise it would explicitly override the TTLs for the `vault.service.consul` DNS record, which I'd call unexpected behavior.

For any other people running into this: you can configure consul to set a TTL on the vault service definition. The rest should be handled automatically. Config snippet:

```json
{
  "dns_config": {
    "service_ttl": {
      "vault": "5s"
    }
  }
}
```

More info here: https://www.consul.io/docs/guides/dns-cache.html

rkrzr:
> In our production setup, we currently cache DNS lookups on the host level.

The title of this issue is a bit misleading. We still don't cache any DNS lookups, what we instead do, is get every secret (which can contain multiple key-value pairs) only once, instead of for every key.

The associated PR is clearer: https://github.com/channable/vaultenv/pull/27

duijf:
> We still don't cache any DNS lookups, what we instead do, is get every secret (which can contain multiple key-value pairs) only once, instead of for every key.

Both statements are true. We do the DNS caching, but dnsmasq takes care of it, not vaultenv :smile:

PR #27, opened 2017-09-19
Make vaultenv only send one request per secret

fatho:
Now, only one request is send even when multiple keys from a secret are
requested. See discussion of issue #18.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 
-import Control.Lens           (preview)
+import Control.Monad          (forM)
+import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Char
 import Data.Either            (isLeft)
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -222,11 +228,11 @@ runCommand options env =
     -- replaces the current process with `command`. It does not return.
     executeFile command searchPath args env'
 
-
-requestSecret :: Options -> Secret -> IO (Either VaultError EnvVar)
-requestSecret opts secret =
+-- | Requests all the data associated with a secret from the vault.
```
rkrzr: Nit: Please use imperative case for comments. I.e. `Request all data associated...`

app/Main.hs
```diff
@@ -235,53 +241,61 @@ requestSecret opts secret =
...
 
     shouldRetry = const $ return . isLeft
-    retryAction _retryStatus = doRequest secret request
+    retryAction _retryStatus = doRequest secretPath request
   in
     Retry.retrying vaultRetryPolicy shouldRetry retryAction
 
-
+-- | Requests all the supplied secrets from the vault, but just once,
+-- even if multiple keys are specified for a single secret.
```
rkrzr: Same as above.

rkrzr: Please also explain that this is an optimization to avoid unnecessary round-trips (and DNS requests).

Issue #11, opened 2017-06-08
Make baseDelay configurable


duijf:
Low priority, until a use-case is found.

See also https://github.com/channable/vaultenv/pull/10#discussion_r120889977

PR #30, opened 2018-01-03
Option for retry-base-delay

yoricksijsling:
Makes the base delay for the vault exponential backoff reconnection
strategy configurable, as requested in #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,12 @@ Available options:
   --secrets-file FILENAME  config file specifying which secrets to request
   --no-connect-tls         don't use TLS when connecting to Vault
   --no-inherit-env         don't merge the current environment with the secret
+  --retry-base-delay-milliseconds MILLISECONDS
+                           base delay for vault connection retrying. Defaults to
+                           40ms because, in testing, we found out that fetching
+                           50 secrets takes roughly 200 milliseconds
+  --retry-attempts NUM     maximum number of vault connection retries. Defaults
+                           to 9
   CMD                      command to run after fetching secrets
   ARGS...                  arguments to pass to CMD, defaults to nothing
   -h,--help                Show this help text
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -43,6 +43,7 @@ data Options = Options
   , oArgs            :: [String]
   , oConnectInsecure :: Bool
   , oInheritEnvOff   :: Bool
+  , oRetryBaseDelay  :: MilliSeconds
   } deriving (Eq, Show)
 
 data Secret = Secret
@@ -70,6 +71,9 @@ data VaultError
   | InvalidUrl        String
   | DuplicateVar      String
   | Unspecified       Int LBS.ByteString
+
+newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
+  deriving (Eq, Show)
 
 --
 -- Argument parsing
@@ -108,6 +112,11 @@ optionsParser env = Options
        <*> switch
            (  long "no-inherit-env"
            <> help "don't merge the parent environment with the secrets file")
+       <*> (MilliSeconds <$> option auto
+               (  long "retry-base-delay"
+               <> metavar "MILLISECONDS"
+               <> value (40 :: Int)
+               <> help "base delay for vault connection retrying. Defaults to 40ms because, in testing, we found out that fetching 50 secrets takes roughly 200 milliseconds"))
   where
     environ vars key = maybe mempty value (lookup key vars)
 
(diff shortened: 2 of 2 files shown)


yoricksijsling:
Difference in delays can be seen when comparing the following (different values for `--retry-base-delay` result in different `real` times):

```
yorick@devm0:~/channable/vaultenv$ time stack exec vaultenv -- --token ABC --secrets-file ~/channable/warpmachine/package/config/warpmachine.secrets --retry-base-delay 10 env
[ERROR] ServerUnreachable error: HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 8200
  secure               = True
  requestHeaders       = [("x-vault-token","ABC")]
  path                 = "/v1/secret/warpmachine/jwt"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HostCannotConnect "localhost" [Network.Socket.connect: <socket: 38>: does not exist (Connection refused)]))

real	0m4.298s
user	0m0.164s
sys	0m0.060s
yorick@devm0:~/channable/vaultenv$ time stack exec vaultenv -- --token ABC --secrets-file ~/channable/warpmachine/package/config/warpmachine.secrets --retry-base-delay 100 env
[ERROR] ServerUnreachable error: HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 8200
  secure               = True
  requestHeaders       = [("x-vault-token","ABC")]
  path                 = "/v1/secret/warpmachine/jwt"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HostCannotConnect "localhost" [Network.Socket.connect: <socket: 38>: does not exist (Connection refused)]))

real	0m28.652s
user	0m0.228s
sys	0m0.144s
```

app/Main.hs
```diff
@@ -108,6 +112,11 @@ optionsParser env = Options
        <*> switch
            (  long "no-inherit-env"
            <> help "don't merge the parent environment with the secrets file")
+       <*> (MilliSeconds <$> option auto
+               (  long "retry-base-delay"
```
ruuda: It would be a good idea to also put the time unit in the flag: somebody who encounters
```
vaultenv --retry-base-delay 100 --token ABC --secrets-file ~/channable/warpmachine/package/config/warpmachine.secrets  env
```
in a shell script or systemd unit file, has no clue whether those are 100 seconds, milliseconds, nanoseconds, or Martian days.
```
vaultenv --retry-base-delay-milliseconds 100 --token ABC --secrets-file ~/channable/warpmachine/package/config/warpmachine.secrets  env
```
on the other hand is self-documenting and leaves no room for ambiguity.

NunoAlexandre: I agree with @ruuda. I deeply dislike having no certainty about the units of values I have to pass to a program.

yoricksijsling: Agreed! I'll do that

app/Main.hs
```diff
@@ -108,6 +112,11 @@ optionsParser env = Options
        <*> switch
            (  long "no-inherit-env"
            <> help "don't merge the parent environment with the secrets file")
+       <*> (MilliSeconds <$> option auto
+               (  long "retry-base-delay-milliseconds"
+               <> metavar "MILLISECONDS"
+               <> value (40 :: Int)
+               <> help "base delay for vault connection retrying. Defaults to 40ms because, in testing, we found out that fetching 50 secrets takes roughly 200 milliseconds"))
```
duijf: Could you also update the [usage section in the README](https://github.com/channable/vaultenv#usage) with the changed output? (Some other options might also be missing)

app/Main.hs
```diff
@@ -121,16 +130,12 @@ optionsInfo env =
 -- | Retry configuration to use for network requests to Vault.
 -- We use a limited exponential backoff with the policy
 -- fullJitterBackoff that comes with the Retry package.
-vaultRetryPolicy :: (MonadIO m) => Retry.RetryPolicyM m
-vaultRetryPolicy =
+vaultRetryPolicy :: (MonadIO m) => MilliSeconds -> Retry.RetryPolicyM m
```
duijf: While we're at it, we can maybe also make `maxRetries` configurable?

yoricksijsling:
Changed the requested @duijf

README.md
```diff
@@ -106,6 +106,12 @@ Available options:
   --secrets-file FILENAME  config file specifying which secrets to request
   --no-connect-tls         don't use TLS when connecting to Vault
   --no-inherit-env         don't merge the current environment with the secret
+  --retry-base-delay-milliseconds MILLISECONDS
+                           base delay for vault connection retrying. Defaults to
+                           40ms because, in testing, we found out that fetching
+                           50 secrets takes roughly 200 milliseconds
```
fatho: Just out of curiosity, how did we reach the conclusion that 40ms is the right base delay based on the fact that 50 secrets take 200 ms?

duijf: I remember deciding this with @ruuda, but I don't know how we came to that number anymore and I can't seem to find the old discussion. (Maybe Ruud knows more details)

I remember it being a bit hand-wavy: "this normally takes 200ms to fetch 50 secrets, so if we wait a bit we can hopefully keep this reasonably snappy without running out of retry options or overloading Vault". (Our internal tooling had around that number of secrets at the time).

duijf: If you have a better way to think up a number, we can reconsider. Then again, it has been working pretty well for us

Issue #34, opened 2018-02-12
Support snakeoil certs


duijf:
People might want to serve a self signed cert for transport layer security. We should add an option to skip cert validation in order to facilitate this

duijf:
This has been implemented.

PR #35, opened 2018-02-12
Add option to skip cert validation

duijf:
This PR adds a new CLI option `--no-validate-certs` to vaultenv. It addresses #34 

Still needs an integration test, but I have verified this locally

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -17,7 +17,6 @@ import Options.Applicative    hiding (Parser, command)
 import System.Environment
 import System.Posix.Process
 import System.IO              (stderr, hPutStrLn)
-import Control.Monad.Reader   (ReaderT, runReaderT, asks)
 import Control.Monad.Except   (ExceptT, MonadError, runExceptT, throwError)
 
 import qualified Control.Concurrent.Async   as Async
@@ -62,7 +61,7 @@ type EnvVar = (String, String)
 
 data Context
   = Context
-  { cEnvVars :: [EnvVar]
+  { cLocalEnvVars :: [EnvVar]
   , cCliOptions :: Options
   , cHttpManager :: Manager
   }
@@ -159,16 +158,16 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds (oRetryBaseDelay
 
 main :: IO ()
 main = do
-  envVars <- getEnvironment
-  cliOptions <- execParser (optionsInfo envVars)
+  localEnvVars <- getEnvironment
+  cliOptions <- execParser (optionsInfo localEnvVars)
   httpManager <- getManager cliOptions
 
-  let context = Context { cEnvVars = envVars
+  let context = Context { cLocalEnvVars = localEnvVars
                         , cCliOptions = cliOptions
                         , cHttpManager = httpManager
                         }
 
-  (runExceptT $ runReaderT vaultEnv context) >>= \case
+  runExceptT (vaultEnv context) >>= \case
     Left err -> hPutStrLn stderr (vaultErrorLogMessage err)
     Right newEnv -> runCommand context newEnv
 
@@ -186,16 +185,11 @@ getManager opts =
   in
     newManager managerSettings
 
-vaultEnv :: ReaderT Context (ExceptT VaultError IO) [EnvVar]
-vaultEnv = do
-  secretFile <- asks (oSecretFile . cCliOptions)
-  secrets <- readSecretList secretFile
-  opts <- asks cCliOptions
-  manager <- asks cHttpManager
-  secretEnv <- requestSecrets manager opts secrets
-  localEnv <- asks cEnvVars
-  inheritEnvOff <- asks (oInheritEnvOff . cCliOptions)
-  checkNoDuplicates (buildEnv localEnv secretEnv inheritEnvOff)
+vaultEnv :: Context -> (ExceptT VaultError IO) [EnvVar]
+vaultEnv context = do
+  secrets <- readSecretList (oSecretFile . cCliOptions $ context)
+  secretEnv <- requestSecrets context secrets
+  checkNoDuplicates (buildEnv secretEnv)
     where
       checkNoDuplicates :: MonadError VaultError m => [EnvVar] -> m [EnvVar]
       checkNoDuplicates e =
@@ -214,9 +208,11 @@ vaultEnv = do
 
       isDup x = foldr (\y acc -> acc || x == y) False
 
-      buildEnv :: [EnvVar] -> [EnvVar] -> Bool -> [EnvVar]
-      buildEnv local remote inheritEnvOff =
-        if inheritEnvOff then remote else remote ++ local
+      buildEnv :: [EnvVar] -> [EnvVar]
+      buildEnv secretsEnv =
+        if (oInheritEnvOff . cCliOptions $ context)
+        then secretsEnv
+        else secretsEnv ++ (cLocalEnvVars context)
 
 
 parseSecret :: String -> Either String Secret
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -55,7 +59,12 @@ data Secret = Secret
 
 type EnvVar = (String, String)
 
-type ExecCtx = ([EnvVar], Options)
+data ExecCtx
```
rkrzr: s/ExecCtx/ExecutionContext

app/Main.hs
```diff
@@ -55,7 +59,12 @@ data Secret = Secret
...
+  = ExecCtx
+  { ctxEnv :: [EnvVar]
+  , ctxOpts :: Options
+  , ctxMan :: Manager
```
rkrzr: I'm not a fan of the abbreviations here (or in general).

I'd go with `contextEnvironment`, `contextOptions`, `contextManager`.
Sure, it's a bit longer, but it's also unambiguous, which is more important imo.

E.g. `ctxMan` - is that the C transaction manual, perhaps?

app/Main.hs
```diff
@@ -148,20 +160,36 @@ main :: IO ()
 main = do
   env <- getEnvironment
   opts <- execParser (optionsInfo env)
+  manager <- getManager opts
 
-  eres <- runExceptT $ runReaderT vaultEnv (env, opts)
+  eres <- runExceptT $ runReaderT vaultEnv (ExecCtx env opts manager)
```
rkrzr: s/eres/eitherNewEnvironment

app/Main.hs
```diff
@@ -148,20 +160,36 @@ main :: IO ()
...
   case eres of
     Left err -> hPutStrLn stderr (vaultErrorLogMessage err)
     Right newEnv -> runCommand opts newEnv
 
+getManager :: Options -> IO Manager
```
rkrzr: Please add a docstring for this function. It would be nice in general to start using Haddock.
Perhaps we can open another issue for that.

Also:
s/getManager/getHttpManager

Even the type doesn't tell you anything: `IO Manager` -> I guess it manages syscalls, perhaps?

app/Main.hs
```diff
@@ -148,20 +160,36 @@ main :: IO ()
...
+getManager opts =
+  let
+    managerSettings = if oConnectInsecure opts
+                      then defaultManagerSettings
```
rkrzr: I'm confused by the logic here: Why are there both `oConnectInsecure` and `oNoValidateCerts`? And why are they mutually exclusive?
I would expect the former to imply the latter.

(Also, I think a `where` would read nicer here)

app/Main.hs
```diff
@@ -255,10 +284,10 @@ requestSecret opts secretPath =
 -- multiple keys are specified for a single secret. This is an optimization in
 -- order to avoid unnecessary round trips and DNS requets.
 requestSecrets :: (MonadError VaultError m, Traversable t, MonadIO m)
-               => Options -> t Secret -> m (t EnvVar)
-requestSecrets opts secrets = do
+               => Manager -> Options -> t Secret -> m (t EnvVar)
+requestSecrets man opts secrets = do
```
rkrzr: s/man/httpManager
s/opts/options

app/Main.hs
```diff
@@ -255,10 +284,10 @@ requestSecret opts secretPath =
...
   let secretPaths = Foldable.foldMap (\x -> Map.singleton x x) $ fmap sPath secrets
-  esecretData <- liftIO $ Async.mapConcurrently (requestSecret opts) secretPaths
+  esecretData <- liftIO $ Async.mapConcurrently (requestSecret man opts) secretPaths
```
rkrzr: s/esecretData/eitherSecretData

app/Main.hs
duijf: `oConnectInsecure` means "connect over HTTP". If we don't connect over TLS, we want the default manager (which doesn't attempt to set up a TLS connection, so we don't need to have any TLS specific settings).

I'll add a comment about this

duijf:
Thanks for the review. I'll do a pass on variable names and comments (most of them weren't mine originally). I've also replied to your comment about the logic, which I'll document

app/Main.hs
duijf: Fixed

app/Main.hs
duijf: Since it is just a temporary binding that we immediately throw into a case and isn't needed any further, I've enabled `LambdaCase`, which works around the problem. In this instance, I'd argue the naming is just noise.

The extension doesn't do much, but it alleviates the need for giving a name to the binding. I'd say it should be in our standard set of extensions to allow/enable. More on [this idiom](http://storm-country.com/blog/LambdaCase)

app/Main.hs
duijf: I went with `secretDataOrErr`

duijf:
So I did a pass on naming, made some type signatures monomorphic (we just always operated on lists instead of the Traversable) and got rid of the `ReaderT`, which wasn't really buying us anything. Could you have another look @rkrzr?

app/Main.hs
```diff
@@ -1,19 +1,26 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
```
duijf: Documented my reasoning for this here: https://github.com/channable/vaultenv/pull/35#discussion_r167576869 (in case you  miss the collapsed comment)

rkrzr: Saving a few characters is not really worth turning on a language extension imo (you probably introduced more chars than you saved with this pragma). But if you prefer it this way, then that's fine with me too.

app/Main.hs
```diff
@@ -1,19 +1,26 @@
...
+import Network.HTTP.Simple    (HttpException(..), Request, Response,
+                               defaultRequest, setRequestHeader, setRequestPort,
+                               setRequestPath, setRequestHost, setRequestManager,
+                               setRequestSecure, httpLBS, getResponseBody,
+                               getResponseStatusCode)
 import Options.Applicative    hiding (Parser, command)
-import System.Environment
-import System.Posix.Process
+import System.Environment     (getEnvironment)
+import System.Posix.Process   (executeFile)
```
rkrzr: 👍 thanks for making all the imports explicit. Much better.

app/Main.hs
```diff
@@ -146,23 +162,39 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds (oRetryBaseDelay
...
+    managerSettings = if oConnectHttp opts
+                      then defaultManagerSettings
+                      else mkManagerSettings tlsSettings Nothing
+    tlsSettings = TLSSettingsSimple
+                { settingDisableCertificateValidation = oNoValidateCerts opts
+                , settingDisableSession = False
+                , settingUseServerName = True
+                }
+
+vaultEnv :: Context -> (ExceptT VaultError IO) [EnvVar]
```
rkrzr: Docstring please.

rkrzr: I would also drop the parentheses around `(ExceptT VaultError IO)`. To me they are more confusing than helpful.

app/Main.hs
duijf: Yeah, it's not really about the net characters saved, but about the overhead of having to come up with names for the temporary bindings (small papercut). When reading the code, this way you also know that the binding will not be used elsewhere. The code becomes "about the logic/dataflow" instead of "about the names given tot things"

`</philosophical-rambling>`

app/Main.hs
duijf: Done. They were left over from when the typesignature was `vaultEnv :: ReaderT ExecCtx (ExceptT VaultError IO) [EnvVar]`

duijf:
All done. Merging.

PR #38, opened 2018-02-21
Configuration via environment variables

### LINKED ISSUE #28: Read vault config from environment?
I don't want to respecify `--host` in every service we have.  Just have some global `EnvironmentFile` that sets the setting.

duijf:
This PR allows us to configure `vaultenv` using environment variables (fixes #28). We support setting the following options:

 - `--host`
 - `--port`
 - `--token`
 - `--secrets-file`
 - `--[no-]connect-tls`
 - `--[no-]validate-certs`
 - `--[no-]inherit-env`
 - `--retry-base-delay-milliseconds`
 - `--retry-attempts`

Notes:

 - The first three are named after the corresponding variables which can be used to configure the official `vault` CLI: `VAULT_HOST`, `VAULT_PORT`, and `VAULT_TOKEN`. 
 - The others are prefixed with `VAULTENV`. 
 - The options which disable defaults are named `VAULTENV_NO_*`, because they explicitly disable the default behavior.
 - CLI arguments take precedence over environment variables.
 - Boolean settings are parsed from the strings `"true"` and `"false"`. No other formats are supported.
 - Because some users might want to override `VAULTENV_NO_*` environment variables, I added in new options: `--validate-certs`, `--inherit-env`, `--connect-tls`, which are no-ops by default, but can override these environment variables. These options are mutually exclusive: vaultenv will throw an error when it encounters both `--connect-tls` and `--no-connect-tls`. (Although the error message isn't that great)

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -95,6 +95,7 @@ main :: IO ()
 main = do
   localEnvVars <- getEnvironment
   cliAndEnvOptions <- parseOptionsFromEnvAndCli localEnvVars
+  print cliAndEnvOptions
 
   httpManager <- getHttpManager cliAndEnvOptions
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -131,18 +131,39 @@ optionsParser envSwitches environment = Options
     <*> many (argument str
       (  metavar "ARGS..."
       <> help "Arguments to pass to CMD, defaults to nothing"))
-    <*> flag (esNoConnectTls envSwitches) True
-      (  long "no-connect-tls"
-      <> help ("Don't use TLS when connecting to Vault. Default: use TLS. Also " ++
-               "configurable via VAULTENV_NO_CONNECT_TLS."))
-    <*> flag (esNoValidateCerts envSwitches) True
-      (  long "no-validate-certs"
-      <> help ("Don't validate TLS certificates when connecting to Vault. Default: " ++
-               "validate certs. Also configurable via VAULTENV_NO_VALIDATE_CERTS."))
-    <*> flag (esNoInheritEnv envSwitches) True
-      (  long "no-inherit-env"
-      <> help ("Don't merge the parent environment with the secrets file. Default: " ++
-               "merge environments. Also configurable via VAULTENV_NO_INHERIT_ENV."))
+    <*>
+      ( flag (esNoConnectTls envSwitches) True
+        (  long "no-connect-tls"
+        <> help ("Don't use TLS when connecting to Vault. Default: use TLS. Also " ++
+                "configurable via VAULTENV_NO_CONNECT_TLS."))
+      <|> flag (esNoConnectTls envSwitches) False
+        (  long "connect-tls"
+        <> help ("Always connect to Vault via TLS. Default: use TLS. Can be used " ++
+                 "to override VAULTENV_NO_CONNECT_TLS.")
+        )
+      )
+    <*>
+      ( flag (esNoValidateCerts envSwitches) True
+        (  long "no-validate-certs"
+        <> help ("Don't validate TLS certificates when connecting to Vault. Default: " ++
+                "validate certs. Also configurable via VAULTENV_NO_VALIDATE_CERTS."))
+      <|> flag (esNoValidateCerts envSwitches) False
+        (  long "validate-certs"
+        <> help ("Always validate TLS certificates when connecting to Vault. Default: " ++
+                 "validate certs. Can be used to override VAULTENV_NO_CONNECT_TLS.")
+        )
+      )
+    <*>
+      ( flag (esNoInheritEnv envSwitches) True
+        (  long "no-inherit-env"
+        <> help ("Don't merge the parent environment with the secrets file. Default: " ++
+                "merge environments. Also configurable via VAULTENV_NO_INHERIT_ENV."))
+      <|> flag (esNoInheritEnv envSwitches) False
+        (  long "inherit-env"
+        <> help ("Always merge the parent environment with the secrets file. Default: " ++
+                 "merge environments. Can be used to override VAULTENV_NO_INHERIT_ENV.")
+        )
+      )
     <*> (MilliSeconds <$> option auto
             (  long "retry-base-delay-milliseconds"
             <> metavar "MILLISECONDS"
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -2,10 +2,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 
---
--- Stackage and stdlib
---
-
 import Control.Monad          (forM)
 import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -2,12 +2,16 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 
+--
+-- Stackage and stdlib
+--
+
 import Control.Monad          (forM)
 import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Char              (toUpper)
 import Data.Either            (isLeft)
-import Data.List              (findIndex, lookup)
+import Data.List              (findIndex)
 import Data.Monoid            ((<>))
 import Network.Connection     (TLSSettings(..))
 import Network.HTTP.Client    (defaultManagerSettings)
@@ -17,7 +21,6 @@ import Network.HTTP.Simple    (HttpException(..), Request, Response,
                                setRequestPath, setRequestHost, setRequestManager,
                                setRequestSecure, httpLBS, getResponseBody,
                                getResponseStatusCode)
-import Options.Applicative    hiding (Parser, command)
 import System.Environment     (getEnvironment)
 import System.Posix.Process   (executeFile)
 import System.IO              (stderr, hPutStrLn)
@@ -35,32 +38,16 @@ import qualified Data.Foldable              as Foldable
 import qualified Data.Map                   as Map
 import qualified Data.Map.Lens              as Lens (toMapOf)
 import qualified Data.Text                  as Text
-import qualified Options.Applicative        as O
-import qualified Text.Read as Read
+
+--
+-- Internal imports
+--
+
+import Config
 
 --
 -- Datatypes
 --
-
-data Options = Options
-  { oVaultHost       :: String
-  , oVaultPort       :: Int
-  , oVaultToken      :: String
-  , oSecretFile      :: FilePath
-  , oCmd             :: String
-  , oArgs            :: [String]
-  , oNoConnectTls    :: Bool
-  , oNoValidateCerts :: Bool
-  , oNoInheritEnv    :: Bool
-  , oRetryBaseDelay  :: MilliSeconds
-  , oRetryAttempts   :: Int
-  } deriving (Eq, Show)
-
-data EnvSwitches = EnvSwitches
-  { esNoConnectTls :: Bool
-  , esNoValidateCerts :: Bool
-  , esNoInheritEnv :: Bool
-  }
 
 data Secret = Secret
   { sPath    :: String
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -93,62 +100,143 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+-- once you know about applicative parsing patterns. We construct a parser for
+-- @Options@ by concatenating parsers for parts of the record.
+--
+-- Toy example to illustrate the pattern:
+--
+-- @
+--     OptionRecord <$> parser1 <*> parser2 <*> parser3
+-- @
+--
+-- Here, the parser for @OptionRecord@ is the combination of parsers of it's
```
rkrzr: s/it's/its

app/Main.hs
```diff
@@ -93,62 +100,143 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
-               <> value (40 :: Int)
-               <> help "base delay for vault connection retrying. Defaults to 40ms because, in testing, we found out that fetching 50 secrets takes roughly 200 milliseconds"))
-       <*> option auto
-           (  long "retry-attempts"
-           <> metavar "NUM"
-           <> value (9 :: Int)
-           <> help "maximum number of vault connection retries. Defaults to 9")
+-- | Parser for our CLI options. Seems intimidating, but is straightforward
+-- once you know about applicative parsing patterns. We construct a parser for
+-- @Options@ by concatenating parsers for parts of the record.
```
rkrzr: s/concatenating/combining

app/Main.hs
```diff
@@ -48,13 +49,19 @@ data Options = Options
...
-  , oConnectHttp     :: Bool
+  , oNoConnectTls    :: Bool
   , oNoValidateCerts :: Bool
-  , oInheritEnvOff   :: Bool
+  , oNoInheritEnv    :: Bool
   , oRetryBaseDelay  :: MilliSeconds
   , oRetryAttempts   :: Int
   } deriving (Eq, Show)
 
+data EnvSwitches = EnvSwitches
```
rkrzr: Can you add a haddock string for this datatype that explains why we need this?
Esp. why we have both `EnvSwitches` and `EnvVar`.

(we should add haddock comments to everything at some point, but let's at least do so for new code for now)

app/Main.hs
```diff
@@ -93,62 +100,143 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+      <> metavar "NUM"
+      <> lookupFromEnvWithDefault "VAULTENV_RETRY_ATTEMPTS" 9
+      <> help "Maximum number of vault connection retries. Defaults to 9")
   where
-    environ vars key = maybe mempty value (lookup key vars)
+    lookupFromEnv key = foldMap value (lookup key environment)
+    lookupFromEnvWithDefault key defVal = maybe (value defVal) value (readFromEnvironment key)
+
+    readFromEnvironment :: Read a => String -> Maybe a
+    readFromEnvironment var = lookup var environment >>= Read.readMaybe
```
rkrzr: This parser has become quite massive. I would break it up into smaller parts:

1. Lift the functions in the `where` clause to the top-level. Add a docstring to them.
2. Put each individual parser on the top-level. Also document them (especially the `flag` and `<|>` story).
3. Combine them with `optionsParser envSwitches environment = Options <$> hostParser <*> portParser ...`
4. Maybe move the parsing functions to a different module?

app/Main.hs
```diff
@@ -93,62 +100,143 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+-- if they're set to anything else than @"true"@ or @"false"@.
+parseEnvSwitches :: [EnvVar] -> EnvSwitches
+parseEnvSwitches vars
+  = EnvSwitches
+  { esNoConnectTls = envSwitch "VAULTENV_NO_CONNECT_TLS"
+  , esNoValidateCerts = envSwitch "VAULTENV_NO_VALIDATE_CERTS"
+  , esNoInheritEnv = envSwitch "VAULTENV_NO_INHERIT_ENV"
+  }
+  where
+    envSwitch key =
```
rkrzr: s/envSwitch/lookupEnvSwitch

app/Main.hs
```diff
@@ -163,25 +251,29 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds (oRetryBaseDelay
 main :: IO ()
 main = do
   localEnvVars <- getEnvironment
-  cliOptions <- execParser (optionsInfo localEnvVars)
-  httpManager <- getHttpManager cliOptions
+
+  cliAndEnvOptions <- execParser (optionsInfo (parseEnvSwitches localEnvVars) localEnvVars)
+
+  print cliAndEnvOptions
```
rkrzr: Should this still be here?

app/Main.hs
```diff
@@ -163,25 +251,29 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds (oRetryBaseDelay
 main :: IO ()
 main = do
   localEnvVars <- getEnvironment
-  cliOptions <- execParser (optionsInfo localEnvVars)
-  httpManager <- getHttpManager cliOptions
+
+  cliAndEnvOptions <- execParser (optionsInfo (parseEnvSwitches localEnvVars) localEnvVars)
```
rkrzr: I think giving a name to the parser, makes it slightly more readable:

```haskell
let argumentParser = optionsInfo (parseEnvSwitches localEnvVars) localEnvVars
cliAndEnvOptions <- execParser argumentParser
```

app/Main.hs
duijf: Fixed

app/Main.hs
duijf: Fixed

duijf:
I believe I addressed all your comments. Some notes:

 - I moved the option handling to a separate `Config` module in `src` and added a library section to `package.yaml`
 - I renamed `EnvSwitches` to `EnvFlags` and documented them. The reason for the `EnvVar` type alias was just readability.
 - I didn't put the individual parsers to the toplevel. I opted to put them in the `where` clause instead. Otherwise I'd have to continue to pass in `envFlags` and `envVars` to most subparsers. This way, they are still part of the scope of `optionsParser`

app/Main.hs
```diff
@@ -2,12 +2,16 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 
+--
+-- Stackage and stdlib
+--
```
rkrzr: Not really necessary imo, but I also don't really mind.

src/Config.hs
```diff
@@ -112,59 +112,67 @@ optionsParserWithInfo envFlags localEnvVars =
...
 optionsParser envFlags envVars = Options
-    <$> strOption
-      (  long "host"
+    <$> host
+    <*> port
+    <*> token
+    <*> secretsFile
+    <*> cmd
+    <*> cmdArgs
+    <*> noConnectTls
```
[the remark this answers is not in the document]
duijf: Sure, will do!

package.yaml
```diff
@@ -19,10 +26,11 @@ executables:
...
       - lens-aeson
+      - mtl
       - optparse-applicative
       - retry
       - text
       - unordered-containers
       - unix
       - utf8-string
-      - mtl
+      - vaultenv
```
rkrzr: I would move all the dependencies here up to the new top-level dependencies section. And only keep `vaultenv` here. It's slightly nicer and also removes duplicates.

src/Config.hs
```diff
@@ -0,0 +1,223 @@
...
+  , oCmd             :: String
+  , oArgs            :: [String]
+  , oNoConnectTls    :: Bool
+  , oNoValidateCerts :: Bool
+  , oNoInheritEnv    :: Bool
+  , oRetryBaseDelay  :: MilliSeconds
+  , oRetryAttempts   :: Int
+  } deriving (Eq, Show)
+
+-- | Behaviour flags that we allow users to set via environment variables.
```
rkrzr: s/Behaviour/Behavior (to be consistent with yourself further down)

src/Config.hs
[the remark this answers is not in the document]
duijf: Done

src/Config.hs
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+-- | Parse program options from the command line and the process environment.
+parseOptionsFromEnvAndCli :: [EnvVar] -> IO Options
+parseOptionsFromEnvAndCli envVars =
+  let envFlags = parseEnvFlags envVars
+      parser = optionsParserWithInfo envFlags envVars
+  in OptParse.execParser parser
+
+-- | Parses behaviour flags from a list of environment variables. If an
```
rkrzr: s/behaviour/behavior

src/Config.hs
```diff
@@ -0,0 +1,223 @@
...
+        Just "false" -> False
+        Nothing -> False
+        _ -> errorWithoutStackTrace $ "[ERROR]: Invalid value for environment variable " ++ key
+
+-- | This function adds metadata to the @Options@ parser so it can be used with
+-- execParser.
+optionsParserWithInfo :: EnvFlags -> [EnvVar] -> OptParse.ParserInfo Options
+optionsParserWithInfo envFlags localEnvVars =
+  OptParse.info
+    (optionsParser envFlags localEnvVars <**> OptParse.helper)
```
rkrzr: I think `OptParse.helper <*> optionsParser envFlags localEnvVars` should probably work as well here, and  `<*>` is better known than `<**>`.

src/Config.hs
```diff
@@ -0,0 +1,223 @@
...
+      <> help ("Config file specifying which secrets to request. Also configurable " ++
+               "via VAULTENV_SECRETS_FILE." )
+    cmd = argument str
+      $  metavar "CMD"
+      <> help "command to run after fetching secrets"
+    cmdArgs =  many $ argument str
+      (  metavar "ARGS..."
+      <> help "Arguments to pass to CMD, defaults to nothing")
+    noConnectTls =
+       flag (efNoConnectTls envFlags) True
```
rkrzr: Could you still add a comment to explain the flag logic here? Where does it make sure that command-line flags override env vars?

src/Config.hs
```diff
@@ -0,0 +1,223 @@
...
+      <> readValueFromEnvWithDefault "VAULTENV_RETRY_BASE_DELAY_MS" 40 envVars
+      <> help ("Base delay for vault connection retrying. Defaults to 40ms. " ++
+                "Also configurable via VAULTENV_RETRY_BASE_DELAY_MS."))
+    retryAttempts = option auto
+      $  long "retry-attempts"
+      <> metavar "NUM"
+      <> readValueFromEnvWithDefault "VAULTENV_RETRY_ATTEMPTS" 9 envVars
+      <> help "Maximum number of vault connection retries. Defaults to 9"
+
+-- ! Attempt to parse an optparse default value modifier from a list of
```
rkrzr: s/-- !/-- |  (everywhere in this module)

app/Main.hs
duijf: Removed

package.yaml
duijf: Good suggestion, done!

src/Config.hs
duijf: Fixed

src/Config.hs
duijf: Agreed & changed. I think this was an artifact of me following an optparse tutorial way in the beginning

src/Config.hs
duijf: I've added some more top level documentation to this function. Not sure if that's the right place, but it is a lot of text and didn't want it inline. Is this better?

src/Config.hs
duijf: Fixed

duijf:
Thanks for the reviews! I've taken your suggestions and the docs should be more clear now. Let me know if there's anything else!

src/Config.hs
```diff
@@ -0,0 +1,251 @@
...
+  ( Options(..)
+  , MilliSeconds(..)
+  , parseOptionsFromEnvAndCli
+  ) where
+
+import Control.Applicative ((<*>), (<|>))
+import Data.Monoid ((<>))
+
+import Options.Applicative (value, long, auto, option, metavar, help, flag,
+                            str, argument, many, strOption, (<**>))
```
rkrzr: The `(<**>)` import can now be removed.

duijf:
Oh yeah, I didn't catch that because the `-Wall` and `-Werror` flags were set for the executable only. Fixed that in the latest commit

PR #39, opened 2018-02-26
Parse bug for string environment variables

duijf:
This fixes a small bug/ergonomics issue where strings passed via environment variables needed to be in Haskell format. From now on, we just use the plain string.

This means users can provide strings like so:

    $ VAULT_HOST="vault.company.com" vaultenv ...

Instead of having to provide them as a Haskell value:

    $ VAULT_HOST='"vault.company.con"' vaultenv ...

This isn't a backwards incompatible change, since the previous behavior never made it into a release.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -5,6 +5,7 @@ module Config
   ) where
 
 import Control.Applicative ((<*>), (<|>))
+import Data.List (intercalate)
 import Data.Monoid ((<>))
 
 import Options.Applicative (value, long, auto, option, metavar, help, flag,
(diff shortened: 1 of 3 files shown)


src/Config.hs
```diff
@@ -228,15 +227,35 @@ optionsParser envFlags envVars = Options
...
+              -> OptParse.Mod f String
+stringFromEnv key envVars = foldMap value $ lookup key envVars
+
+-- | Like @stringFromEnv@, but with a default value.
+stringFromEnvWithDefault :: OptParse.HasValue f
+                         => String
+                         -> String
+                         -> [EnvVar]
+                         -> OptParse.Mod f String
+stringFromEnvWithDefault key defVal envVars
```
rkrzr: s/defVal/defaultValue

duijf:
While testing this, I found myself continuously needing to add debug statements, so I added a debug mode to `vaultenv` as well in the last commit. Can you review that one as well, @rkrzr?

PR #40, opened 2018-02-28
Remove double negative from configuration

duijf:
During #38, we introduced new environment variables through which we allow configuring `vaultenv`. They were named a bit poorly, though. I named them after the option `--no-connect-tls`, which was introduced earlier. This was not the best idea.

This PR fixes that potential papercut by renaming these environment variables to something better. While doing this, I also noticed that the "debug" option would be better if it was called LogLevel.  This is just better naming than "debug on" and "debug off", and also allows for future expansion if we need it.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -92,7 +92,7 @@ main = do
   localEnvVars <- getEnvironment
   cliAndEnvOptions <- parseOptionsFromEnvAndCli localEnvVars
 
-  if oDebug cliAndEnvOptions
+  if (oLogLevel cliAndEnvOptions) >= Info
     then print cliAndEnvOptions
     else pure ()
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -92,7 +92,7 @@ main = do
   localEnvVars <- getEnvironment
   cliAndEnvOptions <- parseOptionsFromEnvAndCli localEnvVars
 
-  if (oLogLevel cliAndEnvOptions) >= Info
+  if (oLogLevel cliAndEnvOptions) <= Info
     then print cliAndEnvOptions
     else pure ()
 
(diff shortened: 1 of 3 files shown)


src/Config.hs
```diff
@@ -47,25 +48,40 @@ instance Show Options where
...
+-- | LogLevel to run vaultenv under. Under @Error@, which is the default, we
+-- will print error messages in error cases. Examples: Vault gives 404s, our
+-- token is invalid, we don't have permissions for a certain path, Vault is
+-- unavailable.
+--
+-- Under @Info@, we print some additional information related to the config.
+data LogLevel
+  = Error
+  | Info
+  deriving (Eq, Ord, Show)
```
rkrzr: Deriving `Ord` does not do the right thing here: 

```
λ> Error > Info
False
```

That's because GHC will just order them alphabetically, I think.
You have to define your own `Ord` instance to get the desired ordering.

app/Main.hs
```diff
@@ -92,7 +92,7 @@ main = do
   localEnvVars <- getEnvironment
   cliAndEnvOptions <- parseOptionsFromEnvAndCli localEnvVars
 
-  if oDebug cliAndEnvOptions
+  if (oLogLevel cliAndEnvOptions) >= Info
```
rkrzr: My intuition with log level "info or higher" would be, that it's "info, warning, error, critical".
But that would not make sense here, since you want to print this if it's "info or debug", right?

In general "higher" log level, ususally means "more serious". E.g. in Python as well: https://docs.python.org/2/library/logging.html#logging-levels

See also my comment about the `Ord` instance below.

src/Config.hs
duijf: Agreed on the order switch! We can actually use the derived `Ord` instance. GHC orders them by the order defined in the sum type. With the new changes:

```haskell
data LogLevel
  = Info
  | Error
  deriving (Eq, Ord)
```

We get:

```ghci
*Main Config> Info < Error
True
*Main Config> Info > Error
False
```

PR #44, opened 2018-05-11
Build using channabuild

duijf:
This is a PR which builds `vaultenv` using an internal build tool currently called Channabuild. This tool is intended to get us consistent build chroots that we can hop in and out of.

This is not intended to be merged yet. This PR is currently to showcase the tech internally (but we're doing all development for `vaultenv` in the open, so that's why you're seeing this). There are plans to open source the build tool behind this, but no ETA on that yet.

If you're reading this and are starting to get worried about us imposing internal tooling: `vaultenv` will always remain buildable using standard Haskell build tools. This will eventually automate more parts of the release process, such as shipping Debian packages, creating GitHub releases and stuff like that.


duijf:
Closing in favour of https://github.com/channable/vaultenv/pull/82

If we ever want to revive this or need some utilities from this, we can always reopen.

PR #46, opened 2018-06-01
Improve integration tests

duijf:
This PR makes some UX and code improvements to running the integration tests. Before these changes, the test script would run, but the output had to be inspected manually.

This PR ports the tests to run under `prove(1)` and to use the test anything protocol, meaning we can just run `./test.sh` and know if something fishy is up (for the things we have coverage for), instead of having to inspect the output manually.


duijf:
Superseded by #47

Issue #45, opened 2018-05-30
Specifying secrets backend


dalekurt:
I have created a new secret on vault called `demo` so all my secrets are stored there. 

However, I do notice that vaultenv is using the path `/v1/secret/demo/my/key/value` to find my key values and not `/v1/demo/my/key/value`. Is there away to configure the backend?

```
$ vault write secret/production/third-party api-key=fecb0f6e97c5b37b3a814107682cf68416f072a8
              ^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^ ^^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             backend          path            key                   value
```

ruuda:
Is it useful (or even possible) to use more than one backend at the same time? If not, I would take the backend from a command-line option or environment variable (possibly with a default value) rather than complicating the config file.

duijf:
> Is it useful (or even possible) to use more than one backend at the same time?

Yes and yes :) 

I think Vault now uses new terminology, the page in the docs is called ["Secret Engines"](https://www.vaultproject.io/docs/secrets/index.html). You can for example have a KV engine (previously called a generic backend), and a Database engine.

The first functions like encrypted key/value storage (like we have running in prod at Channable). The second engine has the same API, but generates credentials on the fly by connecting to Postgres/MySQL and running `CREATE USER` statements. There's another one for SSH.

dalekurt:
@duijf Thank you for getting back to me on this "feature request" I do like the approach of defining the "Secret Engine" (or Backend) in the examples above. That would allow for both static and dynamic key values. At the moment with my own implementation I have dynamic key values for mySQL which query a `database/` path on Vault stored to a JSON file then using jq I can isolate and populate the environment variables. Using your proposed method I would be able to define the path within the `.secrets` file. I like that idea.

duijf:
That's something we'd like to support explicitly without users having to glue these things together

I'm currently implementing this. Will keep the config file roughly like I proposed but:

 - I'll change `BACKEND` to `MOUNT`, as that's the new terminology
 - I don't think I will enforce the newlines
 - I'll introduce `VERSION 2` to get better compatibility guarantees in the future

Implementation note: since the parser is becoming more involved, I'll probably make use of some parser library. It's been a while since I played with parser combinators, so I'm going to have to see what's what.

costimuraru:
👍

[...]

PR #47, opened 2018-06-01
Specify mounts

duijf:
This branch (which is a continuation of `integration-test`) contains a fix for #45 

I've:

 - Moved the secret file parser to it's own file.
 - Added integration tests
 - Changed the HTTP request code so it makes use of mounts
 - Changed the secret file format to support the following syntax:

```
VERSION 2

MOUNT secret
testing#key
testing#otherkey

MOUNT otherbackend
testing2#foo
testing2#bar
```

Some edge cases:

 - We can have `MOUNT secret` two times. Doesn't matter, we just combine everything.
 - Each instruction needs to be on a separate line.
 - If version 1 of the secret file is used, we don't take the mount paths into account when translating from path to the name of the environment variable. If version 2 is used, we do use this. This will mean old files will continue to work as expected, but users can also have `foo` secrets in two mountpoints with the same keys

This does **not** contain breaking changes, as the parser detects the version of the secret format and adjusts behavior accordingly.

TODO:

 - Docs (functions are done, README is left)
 - Be more lenient about whitespace
 - Test a bit further?
 - Release?


src/SecretsFile.hs
```diff
@@ -0,0 +1,113 @@
...
+formatVersionP :: ReadP SecretFileVersion
+formatVersionP = const V2 <$> string "VERSION 2"
+
+-- | Parse a secret block
+--
+-- Exclusive to V2 of the format. A secret block consists of a line describing
+-- the mount location followed by secret specifications.
+secretBlockP :: ReadP [Secret]
+secretBlockP = do
+  mountPath <- string "MOUNT " *> manyTill get (char '\n')
```
rkrzr: This parser is too broad:

```haskell
λ> readP_to_S (manyTill get (char '\n')) "asdf asdfa $%^€\n"
[("asdf asdfa $%^\8364","")]
```

We should check what Vault accepts and restrict it to the same set.

src/SecretsFile.hs
```diff
@@ -0,0 +1,113 @@
...
+-- block specifies which mount point it pertains to.
+secretFileP :: ReadP [Secret]
+secretFileP = do
+  fileVersion <- option V1 (formatVersionP <* skipSpaces)
+  case fileVersion of
+    V1 -> many ((secretP fileVersion "secret") <* skipSpaces) <* eof
+    V2 -> concat <$> many (secretBlockP <* skipSpaces) <* eof
+
+formatVersionP :: ReadP SecretFileVersion
+formatVersionP = const V2 <$> string "VERSION 2"
```
rkrzr: Should `VERSION 2` be on its own line? Do you want to strip leading whitespace as well here?

src/SecretsFile.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | Parse a secret specification line
+--
+-- The version of the fileformat we're parsing determines the way we report
+-- variable information. For V2, the mount point is part of the variable name,
+-- to allow for disambiguation. For V1, this is not needed.
+secretP :: SecretFileVersion -> String -> ReadP Secret
+secretP version mount = do
+  varName <- option Nothing $ Just <$> manyTill get (char '=')
+  (path, key) <- (,) <$> manyTill get (char '#') <*> manyTill get (char '\n')
```
rkrzr: We should restrict the characters we accept here to only the ones that can be used in environment variables.

test.sh
```diff
@@ -0,0 +1,52 @@
...
+cat > ${VAULT_SEEDS_V2} <<EOF
+VERSION 2
+
+MOUNT secret
+testing#key
+testing#otherkey
+
+MOUNT secret
+testing2#foo
+testing2#bar
```
rkrzr: Can you add some more tests for the new file format. E.g. syntax that should not be accepted, and that should return some specific error message. Also, a test for the SOME_NAME=test#key syntax would be good (with whitespace, without, with weird chars etc.)

duijf:
I've addressed your style suggestions. Will look into the parser and test comments later

costimuraru:
👍

src/SecretsFile.hs
duijf: Vault accepts anything, but treats `/` specially

mesaugat:
Could this be merged anytime soon? I need to specify my own mount.

mesaugat:
I've been using the default mount `/secret` right now. I think I can wait until this gets merged. 👍

duijf:
@rkrzr I improved the parser to be more lenient about whitespace. Could you have another look?

@mesaugat Thanks for the patience :)

src/SecretsFile.hs
```diff
@@ -0,0 +1,166 @@
...
+  } deriving (Eq, Show)
+
+data SFVersion
+  = V1
+  | V2
+  deriving (Show)
+
+type Parser = MP.Parsec Void String
+
+data SFError = IOError | ParseError (MP.ParseError (MP.Token String) Void)
```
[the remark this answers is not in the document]
rkrzr: Actually, a more important point:
You are shadowing the built-in `type IOError = GHC.IO.Exception.IOException` here.
That is pretty confusing (esp. since you are also using that type further down in `Either IOError String`), so maybe better call it `SomeIOError` or something like that.

src/SecretsFile.hs
```diff
@@ -0,0 +1,166 @@
...
+parseSecretsFile = MP.parse secretsFileP
+
+-- | SpaceConsumer parser, which is responsible for stripping all whitespace.
+--
+-- Sometimes, we require explicit newlines, therefore, we don't handle those
+-- here.
+whitespace :: Parser ()
+whitespace = MPL.space whitespaceChars lineComment blockComment
+  where
+    whitespaceChars = void $ MP.takeWhile1P (Just "whitespace") (\c -> isSpace c && c /= '\n')
```
rkrzr: What does the `void` throw away here?

src/SecretsFile.hs
```diff
@@ -0,0 +1,166 @@
...
+  pure V2
+
+-- | Parse a secret block
+--
+-- Exclusive to V2 of the format. A secret block consists of a line describing
+-- the mount location followed by secret specifications.
+secretBlockP :: Parser [Secret]
+secretBlockP = do
+  _ <- symbol "MOUNT"
+  mountPath <- lexeme (some MPC.alphaNumChar)
```
rkrzr: Is `alphaNumChar` what Vault supports? Maybe link to their docs here.

src/SecretsFile.hs
duijf: It changes the type from `Parser String` to `Parser ()`. Here, that's the whitespace

rkrzr: You are only parsing `isSpace` here. Isn't there an `isWhitespace` combinator already? Or would that also throw away newlines then? It would be nice to throw away *all* whitespace (including all the crazy unicode chars) except for newlines, I'd say.

src/SecretsFile.hs
```diff
@@ -0,0 +1,166 @@
...
+import Data.Char (toUpper, isSpace)
+import Data.Functor (void)
+import Data.List (intercalate)
+import Data.Void (Void)
+import qualified Text.Megaparsec as MP
+import qualified Text.Megaparsec.Char as MPC
+import qualified Text.Megaparsec.Char.Lexer as MPL
+
+data Secret = Secret
+  { sMount :: String
```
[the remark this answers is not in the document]
duijf: I'll stick with the short prefixes for now, since that's what is used elsewhere

src/SecretsFile.hs
duijf: The builtin ones in Megaparsec indeed all get rid of the newlines, that's why we need our own here.

> It would be nice to throw away all whitespace (including all the crazy unicode chars) except for newlines, I'd say.

That's indeed what this does. `isSpace` comes from `Data.Char` and gives us all that for free :) I'll add a comment

src/SecretsFile.hs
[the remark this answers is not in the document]
duijf: This was actually pretty easy thanks to `displayException`

src/SecretsFile.hs
duijf: This is more restrictive. They seem to support almost anything, except control characters.

I've lifted some restrictions, but we still impose the following:

 - No whitespace (would require some quoting and escaping mechanism)
 - No control characters
 - No `#` and `=` in path components. Allowing this would make the following things ambiguous:

```
FOO=foo#bar
foo#bar#baz
```

app/Main.hs
```diff
@@ -64,8 +53,7 @@ type VaultData = Map.Map String String
 
 data VaultError
```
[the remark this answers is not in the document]
duijf: Done! Good suggestion

src/SecretsFile.hs
```diff
@@ -0,0 +1,166 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+module SecretsFile where
```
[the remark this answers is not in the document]
duijf: Done, as in config.

duijf:
Addressed your suggestions. Can you take another look @rkrzr?

PR #54, opened 2018-10-24
Add option to search PATH

duijf:
Before this patch, Vaultenv used to require an absolute path to the executable it launches. This is a good default for when you use Vaultenv for systemd services.

In local development, however, this is too limiting. Especially if you want to use vaultenv to fetch secrets to fetch for use in local development. Then, PATH searching is a feature you want, because executables live in different locations from machine to machine.

This is a quick patch to add this option to Vaultenv. It introduces a new flag `--use-path`, which enables path search. This flag is turned off by default.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -47,6 +47,7 @@ data Options = Options
   , oRetryBaseDelay  :: MilliSeconds
   , oRetryAttempts   :: Int
   , oLogLevel        :: LogLevel
+  , oUsePath         :: Bool
   } deriving (Eq)
 
 instance Show Options where
@@ -63,6 +64,7 @@ instance Show Options where
     , "Base delay:     " ++ (show . unMilliSeconds $ oRetryBaseDelay opts)
     , "Retry attempts: " ++ (show $ oRetryAttempts opts)
     , "Log-level:      " ++ (show $ oLogLevel opts)
+    , "Use PATH:       " ++ (show $ oUsePath opts)
     ]
 
 -- | Behavior flags that we allow users to set via environment variables.
@@ -73,6 +75,7 @@ data EnvFlags = EnvFlags
   { efConnectTls :: Bool
   , efValidateCerts :: Bool
   , efInheritEnv :: Bool
+  , efUsePath :: Bool
   }
 
 -- | LogLevel to run vaultenv under. Under @Error@, which is the default, we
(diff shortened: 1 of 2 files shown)


src/Config.hs
```diff
@@ -110,6 +113,7 @@ parseEnvFlags envVars
   { efConnectTls = lookupEnvFlag "VAULTENV_CONNECT_TLS"
   , efValidateCerts = lookupEnvFlag "VAULTENV_VALIDATE_CERTS"
   , efInheritEnv = lookupEnvFlag "VAULTENV_INHERIT_ENV"
+  , efSearchPath = lookupEnvFlag "VAULTENV_SEARCH_PATH"
```
rkrzr: So you would set this as `VAULTENV_SEARCH_PATH=true`, if you want it to search for the executable, that `vaultenv` should call, in the local PATH, right?

The name actually doesn't sound like a boolean. It sounds like we're setting a concrete path like e.g.
`VAULTENV_SEARCH_PATH=/some/path'.

OTOH I guess this is already true for all the other env variables we have, so we can also just leave it and stay consistent with how it is already.

src/Config.hs
```diff
@@ -279,6 +284,11 @@ optionsParser envFlags envVars = Options
       <> readValueFromEnvWithDefault "VAULTENV_LOG_LEVEL" Error envVars
       <> help ("Log-level to run vaultenv under. Options: 'error' or 'info'. " ++
                "Defaults to 'error'. Also configurable via VAULTENV_LOG_LEVEL")
+    searchPath
+      =  flag (efInheritEnv envFlags) True
+      $  long "search-path"
+      <> help ("Search PATH for the executable to call. Default: " ++
```
rkrzr: s/for the executable to call/for the executable that vaultenv should call

(I was briefly confused if this was about the `vaultenv` executable itself or about the executable that `vaultenv` will exec, after putting the secrets in the environment)

src/Config.hs
duijf: Oh right, that's true. Named it `VAULTENV_USE_PATH` after our offline discussion

src/Config.hs
duijf: Done

duijf:
Done! I'll merge this and cut a release :)

PR #55, opened 2018-10-24
Increment numbers

mesaugat:
Even though the numbers displayed correctly in preview, the raw file didn't look correct


duijf:
Hi @mesaugat, sorry totally missed this one.

This was actually intentional as to not clutter `git diff`s when adding items in the middle of the list, but I can see that not everyone reads these markdown files when they are rendered and the numbering can be distracting.

Issue #57, opened 2018-10-30
Read a config file from a default location


duijf:
Vaultenv is currently configurable using CLI options and environment variables. It would be nice if we could read standard stuff from a config file without messing with wrapper scripts and the like

PR #58, opened 2018-10-31
config: Search for config files in default places

duijf:
Fixes https://github.com/channable/vaultenv/issues/57

This is an enhancement to make `vaultenv` more useful on devmachines.

Currently users need to specify stuff like `--host`, `--port`, and `--token` manually (or create a wrapper script which does). This changes `vaultenv` to look for for files with the `.env` format in three places:

 - `/etc/vaultenv.conf`
 - `$HOME/.config/vaultenv/vaultenv.conf`
 - `$(cwd)/.env`

New config file precedence in case of duplicate settings:

 - `/etc/vaultenv.conf`
 - `$HOME/.config/vaultenv/vaultenv.conf`
 - `$(cwd)/.env`
 - Env vars which are set
 - CLI options

Later items in the list take precedence

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ CLI flags and environment variable lookups). Vaultenv currently looks for these
 files in the following places:
 
  - `/etc/vaultenv.conf`
- - `$HOME/.vaultenv/config.conf`
+ - `$HOME/.config/vaultenv/vaultenv.conf`
  - `$CWD/.env`
 
 These config files support the exact same syntax as the environment variables
--- a/README.md
+++ b/README.md
@@ -144,7 +144,18 @@ Available options:
 
 ```
 
-## Secret specification
+## Configuration
+
+Vaultenv reads configuration from two types of files:
+
+ - A specification of secrets to fetch.
+ - Configuration options for `vaultenv` itself, mostly connection related.
+
+Decoupling these is useful, because this allows for e.g. changing which secrets
+are fetched on a per project basis, while the connection options stay the same.
+Let's first discuss secrets files.
+
+### Secret specification
 
 There are two versions of this specification format. The first version shipped
 with the initial version of Vaultenv, but doesn't allow users to specify custom
@@ -195,7 +206,41 @@ Vaultenv will make the following environment variables available:
  - `FOOBAR` with the contents of the `foobar` field of the secret at
    `production/third-party`.
 
-## Allowed characters
+### Behavior configuration
+
+Vaultenv supports loading behavior configuration from files (in addition to the
+CLI flags and environment variable lookups). Vaultenv currently looks for these
+files in the following places:
+
+ - `/etc/vaultenv.conf`
+ - `$HOME/.vaultenv/config.conf`
+ - `$CWD/.env`
+
+These config files support the exact same syntax as the environment variables
+that you can use to configure Vaultenv. See the `--help` output for a list of
+what's available.
+
+These files follow the `.env` format (as popularized by [this Ruby
+gem](https://github.com/bkeepers/dotenv)). An example:
+
+```
+# /etc/vaultenv.conf
+# Also: comments are allowed if they start with `#`.
+VAULT_TOKEN="your-vault-token"
+VAULT_PORT="8200"
+VAULTENV_INHERIT_ENV="yes"
+```
+
+This is mostly useful for use on development machines. It allows you to:
+
+ - Set global connection options on a per-machine basis. Useful if you run a
+   Vault instance in your VPN.
+ - Set per-user tokens.
+ - Set per-project secrets files.
+
+All while running Vaultenv without any CLI args.
+
+## Allowed characters in environment variables
 
 We disallow the following in any path to keep the parser and format simple and
 unambiguous:
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -7,6 +7,7 @@ import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bifunctor         (first)
 import Data.Either            (isLeft)
+import Data.List              (nubBy)
 import Data.Monoid            ((<>))
 import Network.Connection     (TLSSettings(..))
 import Network.HTTP.Client    (defaultManagerSettings)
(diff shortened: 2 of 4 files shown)


app/Main.hs
```diff
@@ -84,22 +86,27 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds (oRetryBaseDelay
 main :: IO ()
 main = do
   localEnvVars <- getEnvironment
-  cliAndEnvOptions <- parseOptionsFromEnvAndCli localEnvVars
+  envFileSettings <- readConfigFromEnvFiles
 
-  if (oLogLevel cliAndEnvOptions) <= Info
-    then print cliAndEnvOptions
+  -- Deduplicate, give precedence to set env vars over .env files
+  let envAndEnvFileConfig = nubBy (\(x, _) (y, _) -> x == y) localEnvVars ++ envFileSettings
```
fatho: Do you want to deduplicate the `localEnvVars` and then concatenate, or did you mean to deduplicate after concatenating? In either case, adding explicit parentheses here could help clear things up.

Also, FYI (but the way it's written now is fine and clear), there's `Data.Function.on` which allows you to write the custom equality you wrote as ``(==) `on` fst``.

src/Config.hs
```diff
@@ -332,3 +335,36 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+
+-- | Search for environment files in default locations and load them in order.
+--
+-- This function tries to read the following files in order to obtain
+-- environment configuration. This is implicit behavior and allows the user to
+-- configure vaultenv without setting up environment variables or passing CLI
+-- flags. This is nicer for interactive usage.
+readConfigFromEnvFiles :: IO [(String, String)]
+readConfigFromEnvFiles = do
+  homeDir <- Dir.getHomeDirectory
```
fatho: What happens if the current user doesn't have a home directory?

src/Config.hs
```diff
@@ -332,3 +335,36 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+  cwd <- Dir.getCurrentDirectory
+  let
+    machineConfigFile = "/etc/vaultenv.conf"
+    userConfigFile = homeDir ++ "/.vaultenv/config.conf"
+    cwdConfigFile = cwd ++ "/.env"
+
+  loadMachineConfig <- Dir.doesFileExist machineConfigFile
+  loadUserConfig <- Dir.doesFileExist userConfigFile
+  loadCwdConfig <- Dir.doesFileExist cwdConfigFile
```
fatho: Can `doesFileExist` ever fail? If so, we should guard against those exceptions.

arianvp:
Could we use `.config/vaultenv` as specified in https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html ? This makes it play nice with systemd (user) units

app/Main.hs
duijf: Done

src/Config.hs
duijf: That can indeed fail, (I expected this to just read `$HOME`, but I guess that variable can also be unset and this function also does an existence test).

Changed this to `getXdgDir` and am now catching errors here

README.md
```diff
@@ -195,7 +206,41 @@ Vaultenv will make the following environment variables available:
...
 
-## Allowed characters
+### Behavior configuration
+
+Vaultenv supports loading behavior configuration from files (in addition to the
+CLI flags and environment variable lookups). Vaultenv currently looks for these
+files in the following places:
+
+ - `/etc/vaultenv.conf`
+ - `$HOME/.config/vaultenv/config.conf`
```
rkrzr: Is `config.conf` the name that the standard requires? Otherwise I'd call it `vaultenv.conf`.

README.md
```diff
@@ -195,7 +206,41 @@ Vaultenv will make the following environment variables available:
...
+
+```
+# /etc/vaultenv.conf
+# Also: comments are allowed if they start with `#`.
+VAULT_TOKEN="your-vault-token"
+VAULT_PORT="8200"
+VAULTENV_INHERIT_ENV="yes"
+```
+
+This is mostly useful for use on devmachines. It allows you to:
```
rkrzr: Nit: s/devmachines/development machines (since I'm not sure if everyone uses this terminology)

src/Config.hs
```diff
@@ -332,3 +337,42 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+-- configure vaultenv without setting up environment variables or passing CLI
+-- flags. This is nicer for interactive usage.
+readConfigFromEnvFiles :: IO [(String, String)]
+readConfigFromEnvFiles = do
+  xdgDir <- (Just <$> Dir.getXdgDirectory Dir.XdgConfig "vaultenv")
+    `catchIOError` (const $ pure Nothing)
+  cwd <- Dir.getCurrentDirectory
+  let
+    machineConfigFile = "/etc/vaultenv.conf"
+    userConfigFile :: Maybe FilePath
```
rkrzr: Nit: Newline above please

src/Config.hs
```diff
@@ -332,3 +337,42 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+    userConfigFile = fmap (++ "/config.conf") xdgDir
+    cwdConfigFile = cwd ++ "/.env"
+
+  -- @doesFileExist@ doesn't throw exceptions, it catches @IOError@s and
+  -- returns @False@ if those are encountered.
+  loadMachineConfig <- Dir.doesFileExist machineConfigFile
```
rkrzr: s/loadMachineConfig/machineConfigExists

Also, can you move this down to where it's actually used?

src/Config.hs
```diff
@@ -332,3 +337,42 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+  loadUserConfig <- case userConfigFile of
```
rkrzr: s/loadUserConfig/userConfigExists

src/Config.hs
```diff
@@ -332,3 +337,42 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+     Nothing -> pure False
+     Just fp -> Dir.doesFileExist fp
+  loadCwdConfig <- Dir.doesFileExist cwdConfigFile
```
rkrzr: s/loadCwdConfig/cwdConfigExists

src/Config.hs
```diff
@@ -332,3 +337,42 @@ readValueFromEnvWithDefault :: (Read a, OptParse.HasValue f)
...
+
+  machineConfig <- if loadMachineConfig
+    then DotEnv.parseFile machineConfigFile
+    else pure []
+  userConfig <- if loadUserConfig
+    then DotEnv.parseFile (fromJust userConfigFile) -- safe because of loadUserConfig
```
rkrzr: Nit: In theory there is a race condition here (the file could have been removed inbetween the time you checked that it exists and the time of reading it), but I think we're not going to care about this here.

README.md
duijf: No, not really, was just my preference. I'll go for the name change

src/Config.hs
duijf: Yeah that's true, same for the other ones. Don't think we'll ever see it though

PR #60, opened 2018-11-22
vaultenv no longer supports double underscore in env var names

philwhln:
This is regression we hit with the latest version of vaultenv. This was previously supported and is required by some node.js packages.

Added new double underscore test cases...

```
    SecretFile
      SecretFile.readSecretList
        parses all golden tests succesfully FAILED [1]
        rejects all invalid examples succesfully

    Failures:

      test/SecretFileSpec.hs:17:7:
      1) SecretFile.SecretFile.readSecretList parses all golden tests succesfully
           predicate failed on: [Right [Secret {sMount = "secret", sPath = "foo/bar", sKey = "baz", sVarName = "BAR_BAZ"}],Right [Secret {sMount = "secret", sPath = "foob@ar", sKey = "baz", sVarName = "FOOB@AR_BAZ"},Secret {sMount = "secret", sPath = "fo\1364ob@ar", sKey = "baz", sVarName = "FO\1364OB@AR_BAZ"}],Right [Secret {sMount = "secret", sPath = "stuff/and", sKey = "things", sVarName = "SECRET_STUFF_AND_THINGS"}],Left test/golden/v1.secrets:6:19:
           unexpected '='
           expecting '#' or path component
           ,Right [Secret {sMount = "secret", sPath = "foo", sKey = "bar", sVarName = "SECRET_FOO_BAR"},Secret {sMount = "secret", sPath = "foo/baz", sKey = "bar", sVarName = "BAR"},Secret {sMount = "otherthing", sPath = "foo", sKey = "bar", sVarName = "OTHERTHING_FOO_BAR"},Secret {sMount = "otherthing", sPath = "foo/baz", sKey = "bar", sVarName = "BAR"}],Right [Secret {sMount = "secret", sPath = "VERSIONfoobar", sKey = "foo", sVarName = "VERSIONFOOBAR_FOO"}]]

      To rerun use: --match "/SecretFile/SecretFile.readSecretList/parses all golden tests succesfully/"
```

Fixed double underscore in env var name...

```
    Progress 1/2: vaultenv-0.9.0
    SecretFile
      SecretFile.readSecretList
        parses all golden tests succesfully
        rejects all invalid examples succesfully

    Finished in 0.0075 seconds
    2 examples, 0 failures
```

### DIFF
--- a/src/SecretsFile.hs
+++ b/src/SecretsFile.hs
@@ -197,7 +197,7 @@ secretP version mount = do
 -- Please open a ticket if you require looser restrictions.
 secretVarP :: Parser String
 secretVarP = do
-  var <- intercalate "_" <$> sepBy1 (some MPC.alphaNumChar) (MPC.string "_")
+  var <- intercalate "_" <$> sepBy1 (some MPC.alphaNumChar) (some (MPC.string "_"))
   _ <- symbol "="
   pure var
 
--- a/test/golden/v1.secrets
+++ b/test/golden/v1.secrets
@@ -2,3 +2,5 @@ foo#bar
 foo/bar#baz
 FOO=bar#baz
 BAR=foo/baz#quix
+single_underscore=foo/single#underscore
+double__underscore=foo/double#underscore


duijf:
Hi @philwhln, sorry you're having trouble and thanks for the PR! I hope to look into this today, tomorrow otherwise :)

Out of curiosity, which packages/software use this kind of convention?

philwhln:
Thanks @duijf !

> which packages/software use this kind of convention?

We have a bunch of node.js applications and that are using this convention via https://www.npmjs.com/package/nconf

From that page...

```
  //
  // Can also specify a separator for nested keys (instead of the default ':')
  //
  nconf.env('__');
```

src/SecretsFile.hs
```diff
@@ -197,7 +197,7 @@ secretP version mount = do
 -- Please open a ticket if you require looser restrictions.
 secretVarP :: Parser String
 secretVarP = do
-  var <- intercalate "_" <$> sepBy1 (some MPC.alphaNumChar) (MPC.string "_")
+  var <- intercalate "_" <$> sepBy1 (some MPC.alphaNumChar) (some (MPC.string "_"))
```
duijf: This doesn't actually bind `var` to a correct value. It will parse `vars__with__double__underscores` but that example will just be set in the environment as `vars_with_double_underscores`

```haskell
sepBy1 (some MPC.alphaNumChar) (some (MPC.string "_")) :: Parser [String]
```

This turn `vars__with__double__underscores` into `["vars", "with", "double", "underscores"]` (in a `Parser` context)

```haskell
(intercalate "_" ["vars", "with", "double", "underscores"]) == "vars_with_double_underscores"
```

duijf: Another issue with this parser is that it doesn't accept environment variables with leading underscores, (which is apparently also allowed). It also accepts leading numbers (which isn't allowed)

duijf:
If you want to make corrections, I'd be happy to lend a hand :) If you lost interest in fixing this, I'd also be happy to fix this problem myself. Let me know!

duijf:
I added some commits with test data/integration checks on the [`allowed-env-var-characters`](https://github.com/channable/vaultenv/tree/allowed-env-var-characters) branch. Feel free to make use of that if you're interested :)

philwhln:
> If you want to make corrections, I'd be happy to lend a hand :) If you lost interest in fixing this, I'd also be happy to fix this problem myself. Let me know!

Hi @duijf . Sorry the delayed response. I don't think I'm going to get the bandwidth to do a proper job on this any time soon (plus my knowledge of Haskell is zero), but if I do in the New Year, I'll follow up.

duijf:
Thanks for getting back to me! In that case, I'll close this PR and will develop the fix on https://github.com/channable/vaultenv/pull/63

Thanks for reporting :)

Issue #32, opened 2018-02-09
Only retry on connection related errors


fatho:
Vaultenv should only retry when it cannot establish the connection. When instead it cannot find the requested key, it should exit with an error message immediately.

PR #64, opened 2019-01-02
Only retry on connection errors

duijf:
This is a small enhancement to fix https://github.com/channable/vaultenv/issues/32

After merging this, `vaultenv` will only retry requests when it cannot connect to Vault instead of retry in case of bad tokens or nonexistent secrets.

This constitutes an observable change in behavior. I don't consider this breaking backwards compatibility, as this is an implementation detail.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -6,7 +6,6 @@ import Control.Monad          (forM)
 import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bifunctor         (first)
-import Data.Either            (isLeft)
 import Data.List              (nubBy)
 import Data.Monoid            ((<>))
 import Network.Connection     (TLSSettings(..))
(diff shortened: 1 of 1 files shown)


app/Main.hs
```diff
@@ -187,7 +186,15 @@ requestSecret context secretPath =
...
-    shouldRetry = const $ return . isLeft
+    -- Only retry on connection related failures
+    shouldRetry _retryStatus (Right _) = return False
+    shouldRetry _retryStatus (Left e) =
+      return $ case e of
+        ServerError _ -> True
+        ServerUnavailable _ -> True
+        ServerUnreachable _ -> True
+        Unspecified _ _ -> True
+        _ -> False
```
ruuda: With a wildcard that defaults to no retry, if somebody adds a new connection-related constructor to `VaultError`, that will not cause a compile error, and the behavior will be to not retry. On the other hand, if we list exceptions, then adding a new error where retries would not help, would still result in retries (if we forget to add the match here). And without wildcards the list will be long, but at least everything is explicit.

I’ll leave it up to you which of the three options is most appropriate here.

PR #65, opened 2019-01-02
Add support for key/value engine v2

### LINKED ISSUE #62: Parse secrets version 2 data
I can't seem to get vaultenv to work with key-value secrets engine version 2. vaultenv terminates with error 
> [ERROR] Key KEY not found for path PATH

Inspecting Vault's response for version 1 and version 2, it seems likely that the additional 'data' object is causing the problem.

It seems that secrets on vault's key-value/secret engine format 2 is not parsed correctly. (Or documentation is lacking, so I can't understand how to write the secrets file)

In version 2, data is wrapped as such response.data.data.usernmame, here's an actual example from my audit log:

```
    "response": {
        "data": {
            "data": {
                "password": "hmac-sha256:35765f6bd7eb2658b8681e5359c67e5750f13f4f91ecd59210b6dd48a5171dae",
                "username": "hmac-sha256:10f7d662d7052b68c77549181bfa20ee2f26b60d67dd445cad773e23f82fde8d"
            },
            "metadata": {
                "created_time": "hmac-sha256:e37df48fdde45f74021285c9ae71b00dfb367fcf4b93b3bdbd574e9aa20c50a9",
                "deletion_time": "hmac-sha256:32d8ee3d91fe8a45514f70d2d9d0b0d139a494da8dd87ca9405e28d78713acd6",
                "destroyed": false,
                "version": 1
            }
        }
    },
```

vaultenv gives me error-response: 
[ERROR] Key username not found for path data/X/X

my config is
```
VERSION 2

MOUNT prod
report_connection_user=data/X/X#username
```

duijf:
This is a WIP to fix #62 

I went with the "use heuristics" approach to this problem. The two formats are unambiguous, so we don't need any extra information specified by the user to approach this. Normally, I'd go for "explicit", but this approach means users can use `vaultenv` on a secret backend with the same secrets file before and after an upgrade.

Approach:

 - Drop `lens` and `lens-aeson`. They didn't really make it easier than writing the instances by hand
 - Add a `FromJSON` instance for `VaultData` (which is now `data` instead of `type`) that can parse both

Todo: integration test, agree on heuristic approach.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -2,9 +2,10 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 
+import Control.Applicative    ((<|>))
 import Control.Monad          (forM)
-import Control.Lens           (reindexed, to)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Aeson             (FromJSON, (.:))
 import Data.Bifunctor         (first)
 import Data.List              (nubBy)
 import Data.Monoid            ((<>))
@@ -23,15 +24,13 @@ import Control.Monad.Except   (ExceptT, MonadError, runExceptT, mapExceptT, thro
 import qualified Control.Concurrent.Async   as Async
 import qualified Control.Exception          as Exception
 import qualified Control.Retry              as Retry
-import qualified Data.Aeson.Lens            as Lens (key, members, _String)
+import qualified Data.Aeson                 as Aeson
 import qualified Data.Bifunctor             as Bifunctor
 import qualified Data.ByteString.Char8      as SBS
 import qualified Data.ByteString.Lazy       as LBS hiding (unpack)
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import qualified Data.Foldable              as Foldable
 import qualified Data.Map                   as Map
-import qualified Data.Map.Lens              as Lens (toMapOf)
-import qualified Data.Text                  as Text
 import qualified System.Exit                as Exit
 
 import Config (Options(..), parseOptionsFromEnvAndCli, unMilliSeconds,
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -248,20 +303,16 @@ parseResponse secretPath response =
...
     503 -> Left $ ServerUnavailable responseBody
     _   -> Left $ Unspecified statusCode responseBody
 
 
-parseSuccessResponse :: LBS.ByteString -> VaultData
-parseSuccessResponse responseBody =
-  let
-    getter = Lens.key "data" . reindexed Text.unpack Lens.members . Lens._String . to Text.unpack
-  in
-    Lens.toMapOf getter responseBody
```
rkrzr: Good call, looks better now.

duijf:
I've updated the integration tests to run on KV V2, but there still are some errors. Will investigate what's going on

jaygridley:
Hello guys, could you merge this one?

duijf:
Hey @jaygridley! Thanks for the interest and sorry for the wait. 

I got a bit into the woods with this one because the KV engine V2 has a different URL scheme (not only different response format).

That was the reason the integration tests were failing. I wanted `vaultenv` to auto discover the version of the secret engine instead of making people edit the config file and manually include it. Our internal usecase for this is pretty strong, as we use vaultenv on a lot of servers. Having to edit the config would make a migration from V1 to V2 a lot more difficult.

This required an extra HTTP request to `/sys/mounts`. I just now put this in and have gotten things to the point where the integration tests work. The response from that endpoint has a pretty large schema (and this is something that the `aeson` library isn't great at).

Merging this now and cutting a release.

app/Main.hs
```diff
@@ -105,6 +117,39 @@ instance FromJSON VaultData where
...
+            Aeson.withObject "Options" (\opts -> do
+              version <- opts .: "version"
+              case version of
+                Aeson.String "1" -> pure KV1
+                Aeson.String "2" -> pure KV2
+                _ -> fail "unknown version number") options
+          _ -> fail "expected a KV type"))
+    in
+      Aeson.withObject "MountResp" $ \obj -> do
+        pure . traceShowId $ MountInfo (mapMaybe (\v -> parseMaybe getType v) obj)
```
rkrzr: Should the `traceShowId` still be here?

duijf: Nope. Buy removed that one on master.

Issue #67, opened 2019-04-23
Setup travis-ci as continues integration tool


lpaulmp:
Setup continues integration to run the test suite automatically and after that publish the binaries when a new release is pushed.
This issues would address #61 too.

duijf:
Again, thanks for the PR. I had some issues with the release uploader, but will look into those on Tuesday.

Issue #61, opened 2018-11-23
Automate release process


duijf:
We currently have a bit of automation in the release process, but it isn't great. We make some errors during releases, as evidenced by #51. 

We should fully automate the release process from "git tag" to "binaries available".

PR #68, opened 2019-04-23
Add integration with travis-ci and publish binaries

lpaulmp:
Closes issues #67 # #61

This PR adds the integrations with travis-ci and executes the suite test for all branches when travis-ci detects that a new tag has been pushed runs` stack build `,  `stack test` and create a new release and uploads the binaries built for osx and Linux.
The format will be `vaultenv-v$version-$os`  (vaultenv-v0.9.1-osx) I tested forking the repo and trying by myself

Screen Shot 2019-04-23 at 14 38 34

To create the release, you will need to define two environment variables in travis-ci:
- A user that have permissions over the repo `$GITHUB_USER = lpaulmp`
- Generate a Github token with permissions over the repo `$GITHUB_TOKEN = mytoken1234`.

@duijf I hope you find this useful

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,5 +38,5 @@ after_success:
       stack exec -- cp $(which vaultenv) .
       # Use a user within the organization and generate a github token
       ./github-release release --token=$GITHUB_TOKEN --owner=$GITHUB_USER --repo=vaultenv --title=v$TRAVIS_TAG  --tag=$TRAVIS_TAG
-      ./github-release upload  --token=$GITHUB_TOKEN --owner=$GITHUB_USER --repo=vaultenv --file=vaultenv --tag=$TRAVIS_TAG --name=vautenv-$TRAVIS_TAG-$TRAVIS_OS_NAME-x86_64
+      ./github-release upload  --token=$GITHUB_TOKEN --owner=$GITHUB_USER --repo=vaultenv --file=vaultenv --tag=$TRAVIS_TAG --name=vaultenv-$TRAVIS_TAG-$TRAVIS_OS_NAME-x86_64
     fi;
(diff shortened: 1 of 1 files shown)


lpaulmp:
This is the [build](https://travis-ci.org/lpaulmp/vaultenv/builds/523600270) in travis-ci.

duijf:
Thanks a lot for this PR! This looks very useful.

I have a few questions:

 - What specific OS versions is this actually running on? Can we pin this?
 - What do the binaries need to run (e.g. minimum/specific libc version)?
 - Would there be an easy way to extend this to run the integration tests (maybe against a list of Vault versions)? Might be out of scope here

lpaulmp:
the default OS are `Ubuntu Trusty 14.04` and `macOS 10.13 xcode9.4` these are pretty compatible with most of the Linux and macos machines.
> What specific OS versions is this actually running on? Can we pin this?

Yes, we can pin specific versions adding a matrix but I think this would fine for now so then we can add specific release according to the community needs.
example:
```
matrix:
  include:
    - os: linux
      dist: trusty
    - os: osx
      osx_image: xcode7.2
```

> What do the binaries need to run (e.g. minimum/specific libc version)?

The binaries will need minimum the versions that you find in ubuntu 14.04 (pretty older version) or mac os 10.13. Most of the Linux distros have this version installed in server/desktops so update the os version to the latest ubuntu LTS/macos would lose some compatibility for other users. 

> Would there be an easy way to extend this to run the integration tests

yes, you can extend the scope of the test. You can add more entries in the Travis file to execute the test that you are interested in.

duijf:
Hey! Now that https://github.com/channable/vaultenv/pull/65 is merged, we have a reason to test this one out :)

Thanks a lot for the effort! I'll merge this one and will see if I can get it to work for us!

PR #76, opened 2019-09-06
Added addr support to vault

### LINKED ISSUE #66: Support VAULT_ADDR
Most vault tools seem to support a standardized set of environment variables. More specifically the vault cli checks for VAULT_ADDR rather than 2 separate variables VAULT_HOST and VAUL_PORT. It would be great if vualtenv followed the same convention.

joris-burgers:
Fixes #66.

Added support for VAULT_ADDR to the config files, as well as the command
line arguments. Changed the config parsing process such that everything
is parsed to an Options datatype after which the options are merged. The
addr field is after each separate parsing verified to be correct and the
information is distributed over the UseTLs (HTTP/HTTPS), Host and Port.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -194,7 +194,7 @@ main = do
 
   let envAndEnvFileConfig = nubBy (\(x, _) (y, _) -> x == y) (localEnvVars ++ concat (reverse envFileSettings))
 
-  if True || getOptionsValue "Log level" (oLogLevel cliAndEnvAndEnvFileOptions) <= Info
+  if getOptionsValue "Log level" (oLogLevel cliAndEnvAndEnvFileOptions) <= Info
     then print cliAndEnvAndEnvFileOptions
     else pure ()
   print cliAndEnvAndEnvFileOptions
@@ -304,8 +304,8 @@ vaultEnv context = do
 
       buildEnv :: [EnvVar] -> [EnvVar]
       buildEnv secretsEnv =
-        if (getOptionsValue "CliOptions" . oInheritEnv . cCliOptions $ context)
-        then secretsEnv ++ (cLocalEnvVars context)
+        if getOptionsValue "CliOptions" . oInheritEnv . cCliOptions $ context
+        then secretsEnv ++ cLocalEnvVars context
         else secretsEnv
 
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -177,10 +177,10 @@ data VaultError
 -- fullJitterBackoff that comes with the Retry package.
 vaultRetryPolicy :: (MonadIO m) => Options Validated Completed -> Retry.RetryPolicyM m
 vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds 
-                          (getOptionsValue "retry base delay" $ oRetryBaseDelay opts) * 1000
+                          (getOptionsValue oRetryBaseDelay opts) * 1000
                         )
                      <> Retry.limitRetries (
-                          getOptionsValue "Retry attempts" $ oRetryAttempts opts
+                          getOptionsValue oRetryAttempts opts
                         )
 
 --
@@ -197,7 +197,7 @@ main = do
   let envAndEnvFileConfig = nubBy (\(x, _) (y, _) -> x == y) 
                                   (localEnvVars ++ concat (reverse envFileSettings))
 
-  if getOptionsValue "Log level" (oLogLevel cliAndEnvAndEnvFileOptions) <= Info
+  if getOptionsValue oLogLevel cliAndEnvAndEnvFileOptions <= Info
     then print cliAndEnvAndEnvFileOptions
     else pure ()
 
@@ -219,12 +219,12 @@ main = do
 getHttpManager :: Options Validated Completed -> IO Manager
 getHttpManager opts = newManager managerSettings
   where
-    managerSettings = if getOptionsValue "Connect TLS" $ oConnectTls opts
+    managerSettings = if getOptionsValue oConnectTls opts
                       then mkManagerSettings tlsSettings Nothing
                       else defaultManagerSettings
     tlsSettings = TLSSettingsSimple
                 { settingDisableCertificateValidation = 
-                      not $ getOptionsValue "Validate Certs" $ oValidateCerts opts
+                      not $ getOptionsValue oValidateCerts opts
                 , settingDisableSession = False
                 , settingUseServerName = True
                 }
@@ -244,7 +244,7 @@ vaultEnv context = do
   secretEnv <- requestSecrets context mountInfo secrets
   checkNoDuplicates (buildEnv secretEnv)
     where
-      secretFile = getOptionsValue "Secret file" $ oSecretFile (cCliOptions context)
+      secretFile = getOptionsValue oSecretFile (cCliOptions context)
       checkNoDuplicates :: MonadError VaultError m => [EnvVar] -> m [EnvVar]
       checkNoDuplicates e =
         either (throwError . DuplicateVar) (return . const e) $ dups (map fst e)
@@ -264,7 +264,7 @@ vaultEnv context = do
 
       buildEnv :: [EnvVar] -> [EnvVar]
       buildEnv secretsEnv =
-        if getOptionsValue "CliOptions" . oInheritEnv . cCliOptions $ context
+        if getOptionsValue oInheritEnv . cCliOptions $ context
         then secretsEnv ++ cLocalEnvVars context
         else secretsEnv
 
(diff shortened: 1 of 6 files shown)


src/Config.hs
```diff
@@ -101,40 +271,89 @@ instance Read LogLevel where
   readsPrec _ _ = []
 
 -- | Parse program options from the command line and the process environment.
-parseOptionsFromEnvAndCli :: [EnvVar] -> IO Options
-parseOptionsFromEnvAndCli envVars =
-  let envFlags = parseEnvFlags envVars
-      parser = optionsParserWithInfo envFlags envVars
-  in OptParse.execParser parser
+parseOptionsFromEnvAndCli :: [EnvVar] -> [[EnvVar]] -> IO Options
+parseOptionsFromEnvAndCli localEnvVars envFileSettings =
```
duijf: This function was a historic artefact from having the CLI code use defaults supplied by the environment variables.

I think it makes more sense if we have these functions:

```haskell
parseEnvOptions :: IO Options
parseCliOptions :: IO Options
parseEnvFileOptions :: IO [Options]

-- Can be used in a fold over the results of the previous functions
mergeOptions :: Options -> Options -> Options

validateOptions :: Options -> Either OptionsError Options
```

test/integration/invalid_addr.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+echo "1..5"
+
+stack exec -- vaultenv \
```
duijf: Can you add some comments to these test cases to explain:

 - What is being tested
 - Why it is being tested

app/Main.hs
```diff
@@ -186,15 +191,15 @@ main = do
...
-  cliAndEnvAndEnvFileOptions <- parseOptionsFromEnvAndCli envAndEnvFileConfig
+  let envAndEnvFileConfig = nubBy (\(x, _) (y, _) -> x == y) 
+                                  (localEnvVars ++ concat (reverse envFileSettings))
 
-  if (oLogLevel cliAndEnvAndEnvFileOptions) <= Info
+  if getOptionsValue "Log level" (oLogLevel cliAndEnvAndEnvFileOptions) <= Info
     then print cliAndEnvAndEnvFileOptions
     else pure ()
-
+  print cliAndEnvAndEnvFileOptions
```
duijf: Debug print, can probably go?

src/Config.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE FlexibleInstances
```
duijf: Nit: Generally formatted as `{-# LANGUAGE FlexibleInstances #-}`

src/Config.hs
```diff
@@ -193,10 +412,11 @@ optionsParserWithInfo envFlags localEnvVars =
 -- the default values of the different behaviour switches. So, if an
 -- environment variable is used to configure the TLS option, that value will
 -- always be used, except if it is overridden on the CLI.
-optionsParser :: EnvFlags -> [EnvVar] -> OptParse.Parser Options
-optionsParser envFlags envVars = Options
+optionsParser :: OptParse.Parser Options
+optionsParser = Options
```
duijf: Thanks for refactoring/fixing the type signature here :clap:

src/Config.hs
duijf: Also; it doesn't appear you need this extension for this file

duijf:
Oh and just noticed that there is quite some trailing whitespace here. Please configure your editor to strip trailing whitespace before saving a file :)

src/Config.hs
```diff
@@ -101,40 +271,89 @@ instance Read LogLevel where
...
+  , oInheritEnv     = lookupEnvFlag     "VAULTENV_INHERIT_ENV"
+  , oRetryBaseDelay = MilliSeconds <$> lookupEnvInt      "VAULTENV_RETRY_BASE_DELAY"
+  , oRetryAttempts  = lookupEnvInt      "VAULTENV_RETRY_ATTEMPTS"
+  , oLogLevel       = lookupEnvLogLevel "VAULTENV_LOG_LEVEL"
+  , oUsePath        = lookupEnvFlag     "VAULTENV_USE_PATH"
   }
   where
+    err :: String -> a
+    err key = errorWithoutStackTrace $ "[ERROR]: Invalid value for environment variable " ++ key
+    lookupEnvString key = lookup key envVars
```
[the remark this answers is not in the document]
joris-burgers: This will not work, as these functions use the local variable envVars, which needs to be passed explicitly in such a case. But I added the type signatures and comments.

src/Config.hs
```diff
@@ -209,150 +429,122 @@ optionsParser envFlags envVars = Options
...
-  <> readValueFromEnv key envVars
-
 -- | Search for environment files in default locations and load them in order.
 --
 -- This function tries to read the following files in order to obtain
 -- environment configuration. This is implicit behavior and allows the user to
 -- configure vaultenv without setting up environment variables or passing CLI
 -- flags. This is nicer for interactive usage.
-readConfigFromEnvFiles :: IO [(String, String)]
+readConfigFromEnvFiles :: IO [[(String, String)]]
```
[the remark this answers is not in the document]
joris-burgers: I don't think it should, as this function is also used in Main.hs, where it requires this format for both the parsing of the Options datatype, as well as deciding which variables to export. There might also be items in this the return that have no place in the Options datatype, but are needed for correct operation.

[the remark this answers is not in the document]
duijf: Yes, you're correct. I forgot that we also allow other random stuff in these config files; point rescinded :)

src/Config.hs
```diff
@@ -41,48 +49,210 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+
+isOptionsComplete :: Options -> [OptionsError]
+isOptionsComplete opts = concat
+      [
+        [UnspecifiedValue "Token"       | isNothing (oVaultToken opts)]
+      , [UnspecifiedValue "Command"     | isNothing (oCmd opts)]
+      , [UnspecifiedValue "Secret file" | isNothing (oSecretFile opts)]
+      ]
+
+splitAddress :: String -> (Maybe String, String, String)
```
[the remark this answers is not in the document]
joris-burgers: I could not find a function that does exactly as described (and needed), so I propose we keep this for the moment. I added Haddock documentation, including several examples, as well as a number of example tests and property tests.

src/Config.hs
```diff
@@ -40,50 +47,286 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+-- | Phantom type that indicates that an option is
+-- validated by ```validateCopyAddr```
+data Validated
+
+-- | Phantom type that indicates that an option is
+-- not validated by ```isOptionsComplete```
+data UnValidated
+
+-- | An empty set of options, nothing is specified
+emptyOptions :: Options Validated UnCompleted
```
duijf: Doesn't seem to be used anymore

src/Config.hs
```diff
@@ -40,50 +47,286 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+  | UnknownScheme String -- ^ The scheme of the address is invalid, e.g. ftp://
+  | NonNumericPort String -- ^ The port of the address is not an valid integer
+  | HostPortSchemeAddrMismatch Bool String Int String
+      -- ^ The useTls, host, port and scheme do not match the provided address
+
+
+
+
+instance Show OptionsError where
+  show (UnspecifiedValue s) = "The option " ++ s ++ " is required but not specified"
```
duijf: Sort of related: when we detect mulitple errors (awesome that we detect mulitple errors :clap:), we render them like:

```
[ERROR] [The option Command is required but not specified,The option Secret file is required but not specified]
```

Can we split that:

```
[ERROR] The option Command is required but not specified
[ERROR] The option Secret file is required but not specified
```

test/integration/environment_inherritance.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
```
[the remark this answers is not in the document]
joris-burgers: With or without this flag, the tests will correctly verify. The main difference is that without this flag, the tests will run properly and a failure in them will be caught by the check for their return code. If we were to have the flag, the whole script would fail and prove would not be able to properly tell which test failed, it would simply say that the entire script failed.

src/Config.hs
```diff
@@ -40,50 +47,286 @@ newtype MilliSeconds = MilliSeconds { unMilliSeconds :: Int }
...
+  show (InvalidAddrFormat s) = "The address " ++ s ++ " has an invalid format"
+  show (UnknownScheme s)  = "The address " ++ s ++ " has no recognisable scheme, "
+      ++ " expected http:// or https:// at the beginning of the address."
+  show (NonNumericPort s) = "The port " ++ s ++ " is not a valid int value"
+  show (HostPortSchemeAddrMismatch useTLs host port addr) =
```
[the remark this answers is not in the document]
joris-burgers: I clarified the error message, it now reads:
```
[ERROR] Confliciting configuration values in cli options
Vault address: "http://localhost:42"
Vault host: Unspecified
Vault port: 1234
Use TLS: Unspecified
Hint: This can happen when you provide a `addr` which conflicts with `port`, `host`, or `[no-]connect-tls` in either the environment variables or the CLI.
```
I changed the hint a bit, because the conflicts can also occur in the environment variables of the .env files.

src/Config.hs
joris-burgers: This has been updated.

PR #78, opened 2019-09-30
Add documentation about VAULT_ADDR

### LINKED ISSUE #77: Add documentation for VAULT_ADDR support
The feature described in #66 and developed in #76 needs documentation (what will be accepted, when do configs overwrite one another, etc). We also need to document what happens when two configurations differ (for example, the address `http://localhost:80` and the environment variable `VAULT_PORT=22`)

joris-burgers:
Fixes #77 

Added documentation about the recent changes in the config parsing and
the support for VAULT_ADDR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,13 @@ Use PATH:       True
 In cases where no default nor any value is specified, which is possible for `Token`, `Secret file` and
 `Command`, Vaultenv will give an error that it requires these values to operate.
 
+Vaultenv also supports the `VAULT_ADDR` configuration. In such a case, without specifying separate
+parameters for host, port and whether to use TLS or not, one can specify these values in a single value.
+The address always starts with either `http` or `https://`, followed by a either a DNS name
+or an ip-address. The port is specified at the end of the address, using a `:` to separate the host
+and the port. For example: `https://example.com:42` would create a TLS enabled connection
+to the host `example.com` on port `42`.
+
 Other errors can happen with the address configuration. There are two ways of specifying
 where the source files are located, either via the address of via a combination of
 the port, the host and whether to use TLS. As there are two ways of specifying this,
@@ -293,9 +300,9 @@ being specified in the same configuration, like the same file or both as command
 line options, Vaultenv will not choose either way and will report an error.
 
 In the case they are specified in different configurations, like the address in a file
-and the port in the command line options, the higher presedent (as defined above)
+and the port in the command line options, the higher precedent (as defined above)
 is used for that specific value. In this case, this would result in a host of
-`example.com`, no useage of TLS, due to the `http://` scheme, and a port of `42`.
+`example.com`, no usage of TLS, due to the `http://` scheme, and a port of `42`.
 
 Other errors that can happen during parsing are:
 - A non-numeric port in the address, like `http://localhost:my_port`
--- a/README.md
+++ b/README.md
@@ -245,6 +245,19 @@ VAULT_PORT="8200"
 VAULTENV_INHERIT_ENV="yes"
 ```
 
+All while running Vaultenv without any CLI args.
+
+#### Different levels of configuration
+
+It can happen that conflicting configuration values are send to Vaultenv. An example
+would be a global secret file, which is overwritten by a project specific configuration file.
+The order of precedence (from least precedence to most precedence) is as follows:
+ - `/etc/vaultenv.conf`
+ - `$HOME/.config/vaultenv/vaultenv.conf`
+ - `$CWD/.env`
+ - environment variables
+ - command line options
+
 This is useful on development machines. It allows you to:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+ - `/etc/vaultenv.conf`
+ - `$HOME/.config/vaultenv/vaultenv.conf`
+ - `$CWD/.env`
+ - environment variables
+ - command line options
+
+This means that any command line option that is present would overwrite any other configuration.
+If an option is not specified, the default is used. The defaults are as follows:
+```
+Host:           localhost
```
duijf: It'd be a bit more user friendly if these were the actual strings that the user needs to enter in the config file / on the CLI instead of the pretty printed names.

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
 
 All while running Vaultenv without any CLI args.
 
+It can happen that conflicting configuration values are send to Vaultenv. An example
```
duijf: I would have separate paragraphs for the fact ("Vaultenv supports different levels of config. They have an order of precedence") with the use-case ("For example, machine specific config, project specific config")

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+would be a global secret file, which is overwritten by a project specific configuration file.
+The order of precedent (from least important to most important) is as follows:
```
duijf: Typo: `s/predecent/precedence/`. Maybe also `s/important/precedence/`

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+Inherit env:    True
+Base delay:     40
+Retry attempts: 9
+Log-level:      Error
+Use PATH:       True
+```
+In cases where no default nor any value is specified, which is possible for `Token`, `Secret file` and
+`Command`, Vaultenv will give an error that it requires these values to operate.
+
+Vaultenv also supports the `VAULT_ADDR` configuration. In such a case, without specifying separate
```
duijf: This is starting to become quite a wall of text. 

Can we make this its own section "Connection options"? We could add a general introduction that there are two ways to specify connection options: `VAULT_ADDR` and the specific values. Then the rest of the text is a bit easier to read

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+parameters for host, port and whether to use TLS or not, one can specify these values in a single value.
+The address always starts with either `http://` or `https://`, followed by a either a DNS name
+or an ip-address. The port is specified at the end of the address, using a `:` to separate the host
+and the port. For example: `https://example.com:42` would create a TLS enabled connection
+to the host `example.com` on port `42`.
+
+Other errors can happen with the address configuration. There are two ways of specifying
```
duijf: Errors other than what?

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+where the secret files are located, either via the address of via a combination of
```
duijf: s/where the secret files are located/what the connection options are/

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+the port, the host and whether to use TLS. As there are two ways of specifying this,
+it is also possible for these values to conflict. Consider the situation of where
```
duijf: Nit: remove `of`

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+`VAULT_ADDR` is `http://example.com:8200` and `VAULT_PORT` is set to `42`. There
+are two ways Vaultenv can resolve this. In the case of the address and the port
+being specified in the same configuration, like the same file or both as command
+line options, Vaultenv will not choose either way and will report an error.
+
+In the case they are specified in different configurations, like the address in a file
+and the port in the command line options, the higher precedent (as defined above)
```
duijf: `s/precedent/precedence`. Maybe make `above` a link?

README.md
```diff
@@ -244,6 +252,60 @@ This is useful on development machines. It allows you to:
...
+Other errors can happen with the address configuration. There are two ways of specifying
+where the secret files are located, either via the address of via a combination of
+the port, the host and whether to use TLS. As there are two ways of specifying this,
+it is also possible for these values to conflict. Consider the situation of where
+`VAULT_ADDR` is `http://example.com:8200` and `VAULT_PORT` is set to `42`. There
+are two ways Vaultenv can resolve this. In the case of the address and the port
+being specified in the same configuration, like the same file or both as command
+line options, Vaultenv will not choose either way and will report an error.
+
+In the case they are specified in different configurations, like the address in a file
```
duijf: "Configurations" is a bit ambiguous. Should we call it a "configuration level"?

PR #84, opened 2020-01-13
Add option to remove environment variables before executing program

### LINKED ISSUE #83: Support blacklisting of a list of environment variables
Vaultenv currently passes on either *all* env variables (`--inherit-env`) or none (`--no-inherit-env`).

It would be nice to extend this by allowing to blacklist specific env variables by name. This would for example be useful for Vault authentication information, which must be available to Vaultenv, but usually should not be available to processes started by Vaultenv.

maartenberg:
This PR adds the option `--inherit-env-blacklist` to Vaultenv.

This option takes a comma-separated list of environment variable names, and when
defined these environment variables are removed from the environment before Vaultenv
executes the command to run.

This option does not implicitly enable `--inherit-env`, and the blacklist is not
applied if `--inherit-env` is `false` (we could change `buildEnv` so it does if this
is desired).
No warning is given if a variable defined in the secrets file is removed by this
blacklist.

Example usage:

```console
$ stack run -- --secrets-file /tmp/dummy-file env | grep VAULT_
VAULT_TOKEN=present-in-environment
VAULT_HOST=127.0.0.1
VAULT_PORT=8200

$ # Filter Vault token from environment
$ stack run -- --secrets-file /tmp/dummy-file --inherit-env-blacklist VAULT_TOKEN,VAULT_HOST env | grep VAULT_
VAULT_PORT=8200

$ # Equivalent (but does cause the blacklist to end up in the environment):
$ VAULTENV_INHERIT_ENV_BLACKLIST=VAULT_TOKEN,VAULT_HOST stack run -- --secrets-file /tmp/dummy-file env | grep VAULT_
VAULTENV_INHERIT_ENV_BLACKLIST=VAULT_TOKEN,VAULT_HOST
VAULT_PORT=8200
```

Fixes #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,12 +99,13 @@ of `tutorial.sh`.
 ## Usage
 
 ```
-vaultenv - run programs with secrets from HashiCorp Vault
+vaultenv 0.11.0 - run programs with secrets from HashiCorp Vault
 
 Usage: vaultenv [--version] [--host HOST] [--port PORT] [--addr ADDR]
                 [--token TOKEN] [--secrets-file FILENAME] [CMD] [ARGS...]
                 ([--no-connect-tls] | [--connect-tls]) ([--no-validate-certs] |
                 [--validate-certs]) ([--no-inherit-env] | [--inherit-env])
+                [--inherit-env-blacklist COMMA_SEPARATED_NAMES]
                 [--retry-base-delay-milliseconds MILLISECONDS]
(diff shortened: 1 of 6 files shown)


maartenberg:
> Can you update the docs in the README to mention this option?

Done.

> Can you add tests for this behavior and for the three different ways you can pass
> this option to Vaultenv?

Done.

PR #86, opened 2020-01-15
Build .deb package with static binary from Nix

maartenberg:
This PR brings back the scripts to build a .deb package for Vaultenv, after it was
removed in 9fb4125. The scripts have been altered to use the static binary built by
Nix, which means that we can publish the .deb without any dependencies and it should
work on any Debian / Ubuntu derivative.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -6,6 +6,7 @@ in
     paths = [
       pkgs.stack
       pkgs.vault
+	  pkgs.cachix
     ];
   }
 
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -6,6 +6,7 @@ in
     paths = [
       pkgs.stack
       pkgs.vault
+	  pkgs.cachix
```
ruuda: This line contains a tab.

maartenberg: :cry:

PR #89, opened 2020-02-11
Fix retry behavior

### LINKED ISSUE #88: Retries are broken when Vault server is unreachable
We've noticed an issue in production where Vaultenv v0.12.0 does not respect the
retry policy, but instead immediately crashes with what appears to be an unhandled
`HttpExceptionRequest` exception and a message like this one:

```
vaultenv: HttpExceptionRequest Request {
  host                 = "vault.server.address"
  port                 = 1234
  secure               = True
  requestHeaders       = [("x-vault-token","*")]
  path                 = "/v1/secret/foobar"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 ConnectionTimeout
```

We have verified that this issue is not present in Vaultenv v0.9.0 (the previous
version used on this host). Vaultenv v0.9.0 gives a slightly different error message
and does not crash instantly, which seems to indicate that the retries are working in
this version. The message given by Vaultenv 0.9.0 looks like this:

```
[ERROR] ServerUnreachable error: HttpExceptionRequest Request {
  host                 = "vault.server.address"
  port                 = 1234
  secure               = True
  requestHeaders       = [("x-vault-token","*")]
  path                 = "/v1/secret/foobar"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 ConnectionTimeout
```

### LINKED ISSUE #90: More granular retrying of secret retrieval
When Vaultenv fails to fetch some secrets from Vault the current behaviour is to
retry fetching all secrets, including the secrets that were already fetched.

A better solution would be to retry fetching only those secrets that could not be
retrieved, as the current behaviour could (theoretically) cause a Vault server to be
overloaded for longer than without this behaviour. This should not be a noticeable
issue as the full jitter backoff should ensure that the requests don't get retried at
the exact same time.

This behaviour would probably need to be added to `requestSecrets` in the Main module.

maartenberg:
This PR reworks the internal logic of Vaultenv to remove the use of the `ExceptT`
monad and fix the retrying logic for fetching both the mount information and the
secrets.

It was discovered that the existing retry code had no effect due to the `http`
library throwing IO exceptions, which were not caught by the ExceptT monad
combination we used. Additionally, the retries were not enabled on the first call to
the server for the secret store mount information.

In this PR the relevant functions that use the ExceptT construction are reworked or
replaced to return either `Either VaultError result` or `IO (Either VaultError
result)`, and `Retry.retrying` is used directly to retry when a network-related
`VaultError` is returned.

This PR also alters the handling for HttpExceptions to ensure that the Vault root
token is removed before the exception can be shown.

Fixes #88.
Fixes #90.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -173,6 +173,21 @@ data VaultError
   | Unspecified Int LBS.ByteString
   deriving Show
 
+-- | "Handle" a HttpException by wrapping it in a Left VaultError.
+-- We also edit the Request contained in the exception to remove the
+-- Vault token, as it would otherwise get printed to stderr if we error
+-- out.
+httpErrorHandler :: HttpException -> IO (Either VaultError b)
+httpErrorHandler (e :: HttpException) = case e of
+  (HttpExceptionRequest request reason) ->
+    let sanitizedRequest = sanitizeRequest request
+    in pure $ Left $ ServerUnreachable (HttpExceptionRequest sanitizedRequest reason)
+  (InvalidUrlException url _reason) -> pure $ Left $ InvalidUrl url
+
+  where
+    sanitizeRequest :: Request -> Request
+    sanitizeRequest = setRequestHeader "x-vault-token" ["**removed**"]
+
 -- | Retry configuration to use for network requests to Vault.
 -- We use a limited exponential backoff with the policy
 -- fullJitterBackoff that comes with the Retry package.
@@ -183,6 +198,39 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds
                      <> Retry.limitRetries (
                           getOptionsValue oRetryAttempts opts
                         )
+
+-- | Perform the given action according to our retry policy.
+doWithRetries :: Retry.RetryPolicyM IO -> (Retry.RetryStatus -> IO (Either VaultError a)) -> IO (Either VaultError a)
+doWithRetries retryPolicy = Retry.retrying retryPolicy isRetryableFailure
+  where
+    -- | Indicator function for retrying to retry on VaultErrors (Lefts) that
+    -- shouldRetry thinks we should retry on. Needs to be in IO because the
+    -- actions to perform are in IO as well.
+    isRetryableFailure :: Retry.RetryStatus -> Either VaultError a -> IO Bool
+    isRetryableFailure _retryStatus (Right _) = pure False
+    isRetryableFailure retryStatus (Left err) = shouldRetry retryStatus err
+
+    -- | Determine whether a request that resulted in the given VaultError
+    -- should be retried.
+    shouldRetry :: Applicative f => Retry.RetryStatus -> VaultError -> f Bool
+    shouldRetry _retryStatus res = pure $ case res of
+      ServerError _ -> True
+      ServerUnavailable _ -> True
+      ServerUnreachable _ -> True
+      Unspecified _ _ -> True
+      BadJSONResp _ -> True
+
+      -- Errors where we don't retry
+      BadRequest _ -> False
+      Forbidden -> False
+      InvalidUrl _ -> False
+      SecretNotFound _ -> False
+
+      -- Errors that cannot occur at this point, but we list for
+      -- exhaustiveness checking.
+      KeyNotFound _ -> False
+      DuplicateVar _ -> False
+      SecretFileError _ -> False
 
 --
 -- IO
@@ -243,7 +291,7 @@ getHttpManager opts = newManager managerSettings
 -- Signals failure through a value of type VaultError.
 vaultEnv :: Context -> IO (Either VaultError [EnvVar])
 vaultEnv context = do
-  mountInfo <- doWithRetries getMountInfo
+  mountInfo <- doWithRetries retryPolicy getMountInfo
   case mountInfo of
     Left vaultError -> pure $ Left vaultError
     Right mountInfo' -> do
@@ -251,52 +299,15 @@ vaultEnv context = do
       case secrets of
         Left sfError -> pure $ Left $ SecretFileError sfError
         Right secrets' -> do
-          -- NOTE: This retries fetching all the secrets if fetching one of the
-          -- secrets fails due to an error we should retry on. We would ideally
-          -- keep the secrets that were successfully fetched in memory and only
-          -- retry getting the ones we couldn't get yet, as this makes it less
-          -- likely to get a 'thundering herd' of Vaultenv processes all trying
-          -- and failing to get all of their secrets.
-          -- This would then need to happen in requestSecrets, but I don't want
-          -- to touch that right now.
-          -- Issue: https://github.com/channable/vaultenv/issues/90
-          secretEnv <- doWithRetries (getSecrets mountInfo' secrets')
+          secretEnv <- requestSecrets context mountInfo' secrets'
           case secretEnv of
             Left vaultError -> pure $ Left vaultError
             Right secretEnv' -> pure $ checkNoDuplicates (buildEnv secretEnv')
     where
-      doWithRetries :: (Retry.RetryStatus -> IO (Either VaultError a)) -> IO (Either VaultError a)
-      doWithRetries = Retry.retrying retryPolicy isRetryableFailure
-
-      -- | Indicator function for retrying to retry on VaultErrors (Lefts) that
-      -- shouldRetry thinks we should retry on. Needs to be in IO because the
-      -- actions to perform are in IO as well.
-      isRetryableFailure :: Retry.RetryStatus -> Either VaultError a -> IO Bool
-      isRetryableFailure _retryStatus (Right _) = pure False
-      isRetryableFailure retryStatus (Left err) = shouldRetry retryStatus err
-
       retryPolicy = vaultRetryPolicy (cCliOptions context)
 
       getMountInfo :: Retry.RetryStatus -> IO (Either VaultError MountInfo)
       getMountInfo _retryStatus = catch (requestMountInfo context) httpErrorHandler
-
-      getSecrets :: MountInfo -> [Secret] -> Retry.RetryStatus -> IO (Either VaultError [EnvVar])
-      getSecrets mountInfo secrets _retryStatus = catch (requestSecrets context mountInfo secrets) httpErrorHandler
-
-      -- | "Handle" a HttpException by wrapping it in a Left VaultError.
-      -- We also edit the Request contained in the exception to remove the
-      -- Vault token, as it would otherwise get printed to stderr if we error
-      -- out.
-      httpErrorHandler :: HttpException -> IO (Either VaultError b)
-      httpErrorHandler (e :: HttpException) = case e of
-        (HttpExceptionRequest request reason) ->
-          let sanitizedRequest = sanitizeRequest request
-          in pure $ Left $ ServerUnreachable (HttpExceptionRequest sanitizedRequest reason)
-        (InvalidUrlException url _reason) -> pure $ Left $ InvalidUrl url
-
-        where
-          sanitizeRequest :: Request -> Request
-          sanitizeRequest = setRequestHeader "x-vault-token" ["**removed**"]
 
       secretFile = getOptionsValue oSecretFile (cCliOptions context)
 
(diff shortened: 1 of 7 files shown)


app/Main.hs
```diff
@@ -237,17 +239,66 @@ getHttpManager opts = newManager managerSettings
...
 -- Signals failure through a value of type VaultError, but can also
 -- throw HTTP exceptions.
-vaultEnv :: Context -> ExceptT VaultError IO [EnvVar]
+vaultEnv :: Context -> IO (Either VaultError [EnvVar])
 vaultEnv context = do
-  mountInfo <- requestMountInfo context
-  secrets <- mapExceptT (fmap $ first SecretFileError) $ readSecretList secretFile
-  secretEnv <- requestSecrets context mountInfo secrets
-  checkNoDuplicates (buildEnv secretEnv)
+  mountInfo <- doWithRetries getMountInfo
```
rkrzr: I would add a comment here to explain *why* we need to get mounts first.

app/Main.hs
```diff
@@ -237,17 +239,66 @@ getHttpManager opts = newManager managerSettings
 --
 -- Signals failure through a value of type VaultError, but can also
 -- throw HTTP exceptions.
```
rkrzr: The comment is now out of date, since this should never throw HTTP exceptions now.

app/Main.hs
```diff
@@ -237,17 +239,66 @@ getHttpManager opts = newManager managerSettings
...
+        Left sfError -> return $ Left $ SecretFileError sfError
+        Right secrets' -> do
+          -- NOTE: This retries fetching all the secrets if fetching one of the
+          -- secrets fails due to an error we should retry on. We would ideally
+          -- keep the secrets that were successfully fetched in memory and only
+          -- retry getting the ones we couldn't get yet, as this makes it less
+          -- likely to get a 'thundering herd' of Vaultenv processes all trying
+          -- and failing to get all of their secrets.
+          -- This would then need to happen in requestSecrets, but I don't want
+          -- to touch that right now.
```
rkrzr: Let's file a new issue for this.

app/Main.hs
```diff
@@ -237,17 +239,66 @@ getHttpManager opts = newManager managerSettings
...
+      -- shouldRetry thinks we should retry on. Needs to be in IO because the
+      -- actions to perform are in IO as well.
+      isRetryableFailure :: Retry.RetryStatus -> Either VaultError a -> IO Bool
+      isRetryableFailure _retryStatus (Right _) = pure False
+      isRetryableFailure retryStatus (Left err) = shouldRetry retryStatus err
+
+      retryPolicy = vaultRetryPolicy (cCliOptions context)
+
+      getMountInfo _retryStatus = catch (requestMountInfo context) httpErrorHandler
+      getSecrets mountInfo secrets _retryStatus = catch (requestSecrets context mountInfo secrets) httpErrorHandler
```
rkrzr: Please add type signatures for both of these.

rkrzr:
About dropping `mtl`. As a first step we could also just replace it with `transformers`. It's almost a drop-in replacement:

```diff
diff --git a/package.yaml b/package.yaml
index 04e16fa..e4e83df 100644
--- a/package.yaml
+++ b/package.yaml
@@ -18,11 +18,11 @@ dependencies:
   - http-conduit
   - http-client
   - megaparsec
-  - mtl
   - optparse-applicative
   - parser-combinators
   - retry
   - text
+  - transformers
   - unordered-containers
   - unix
   - utf8-string
diff --git a/src/Config.hs b/src/Config.hs
index 755571d..ea63a8d 100644
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -21,7 +21,7 @@ module Config
 
 import Control.Applicative ((<*>), (<|>))
 import Control.Monad(when, unless)
-import Control.Monad.Except (runExcept, throwError)
+import Control.Monad.Trans.Except (runExcept, throwE)
 import Data.List (intercalate, isPrefixOf)
 import Data.Maybe (fromJust, fromMaybe, isNothing, isJust)
 import Data.Monoid ((<>))
@@ -194,9 +194,9 @@ validateCopyAddr source opts
                 (oVaultAddr opts)
         (mStrScheme, addrHost, addrStrPort) = splitAddress addr
     unless (all isDigit addrStrPort && not (null addrStrPort))
-        (throwError $ NonNumericPort addrStrPort)
+        (throwE $ NonNumericPort addrStrPort)
     unless (isJust mStrScheme)
-        (throwError $ UnknownScheme addrHost)
+        (throwE $ UnknownScheme addrHost)
     let mHost = oVaultHost opts
         mPort = oVaultPort opts
         mUseTLS = oConnectTls opts
@@ -211,7 +211,7 @@ validateCopyAddr source opts
             ||
             (isJust mUseTLS && addrTLS /= mUseTLS) -- Is the UseTLS set and the same
     when doesAddrDiffer
-      (throwError $ HostPortSchemeAddrMismatch
+      (throwE $ HostPortSchemeAddrMismatch
           source
           mUseTLS
           mHost
```

But dropping it altogether, would probably be best.

app/Main.hs
maartenberg: Done, #90. Link added to this comment as well.

app/Main.hs
maartenberg: Done.

app/Main.hs
maartenberg: Fixed.

app/Main.hs
maartenberg: Added.

app/Main.hs
```diff
@@ -385,31 +425,27 @@ lookupSecrets mountInfo secrets vaultData = forM secrets $ \secret ->
...
-doRequest :: String -> Request -> ExceptT VaultError IO VaultData
+doRequest :: String -> Request -> IO (Either VaultError VaultData)
 doRequest secretPath request = do
-  resp <- withExceptT exToErr (httpLBS request)
-  parseResponse secretPath resp
-  where
-    exToErr :: HttpException -> VaultError
-    exToErr e@(HttpExceptionRequest _ _) = ServerUnreachable e
-    exToErr (InvalidUrlException _ _) = InvalidUrl secretPath
+  resp <- httpLBS request
```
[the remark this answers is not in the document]
maartenberg: Done.

app/Main.hs
```diff
@@ -305,11 +361,15 @@ requestMountInfo context =
               (getOptionsValue  oConnectTls cliOptions)
         $ defaultRequest
   in do
-    resp <- withExceptT ServerUnreachable (httpLBS request)
-    withExceptT BadJSONResp (liftEither $ Aeson.eitherDecode' (getResponseBody resp))
+    resp <- httpLBS request
```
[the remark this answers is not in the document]
maartenberg: Done.

app/Main.hs
```diff
@@ -305,11 +361,15 @@ requestMountInfo context =
...
+    let decodeResult = Aeson.eitherDecode' (getResponseBody resp) :: Either String MountInfo
+
+    case decodeResult of
+      Left errorMsg -> return $ Left $ BadJSONResp errorMsg
+      Right result -> return $ Right result
```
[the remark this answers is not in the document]
maartenberg: Replaced the new usages of `return` with `pure`.

maartenberg:
> We could also consider dropping our dependency on `mtl` altogether.

I've been able to remove MTL with minimal changes, see d80b1e6.

app/Main.hs
```diff
@@ -229,25 +231,81 @@ getHttpManager opts = newManager managerSettings
...
+      getMountInfo _retryStatus = catch (requestMountInfo context) httpErrorHandler
+
+      getSecrets :: MountInfo -> [Secret] -> Retry.RetryStatus -> IO (Either VaultError [EnvVar])
+      getSecrets mountInfo secrets _retryStatus = catch (requestSecrets context mountInfo secrets) httpErrorHandler
+
+      -- | "Handle" a HttpException by wrapping it in a Left VaultError.
+      -- We also edit the Request contained in the exception to remove the
+      -- Vault token, as it would otherwise get printed to stderr if we error
+      -- out.
+      httpErrorHandler (e :: HttpException) = case e of
```
duijf: Please add a type signature for this binding.

maartenberg: Done.

app/Main.hs
```diff
@@ -208,7 +209,8 @@ main = do
                         , cHttpManager = httpManager
                         }
 
-  runExceptT (vaultEnv context) >>= \case
+  envVars <- vaultEnv context
+  case envVars of
```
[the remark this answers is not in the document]
maartenberg: Nice, added

maartenberg:
@duijf: I've now added a simple integration test for the retry behaviour and tested that it doesn't pass when it is run on the binary on current `master`.

app/Main.hs
```diff
@@ -183,6 +199,39 @@ vaultRetryPolicy opts = Retry.fullJitterBackoff (unMilliSeconds
...
+-- | Perform the given action according to our retry policy.
+doWithRetries :: Retry.RetryPolicyM IO -> (Retry.RetryStatus -> IO (Either VaultError a)) -> IO (Either VaultError a)
+doWithRetries retryPolicy = Retry.retrying retryPolicy isRetryableFailure
+  where
+    -- | Indicator function for retrying to retry on VaultErrors (Lefts) that
+    -- shouldRetry thinks we should retry on. Needs to be in IO because the
+    -- actions to perform are in IO as well.
+    isRetryableFailure :: Retry.RetryStatus -> Either VaultError a -> IO Bool
+    isRetryableFailure _retryStatus (Right _) = pure False
+    isRetryableFailure retryStatus (Left err) = shouldRetry retryStatus err
```
duijf: We can simplify this a bit further (we don't need the separate `shouldRetry` with that general type singature)

```haskell
    isRetryableFailure :: Retry.RetryStatus -> Either VaultError a -> IO Bool
    isRetryableFailure _retryStatus (Right _) = pure False
    isRetryableFailure _retryStatus (Left err) = pure $ case err of
      ServerError _ -> True
      ServerUnavailable _ -> True
      ServerUnreachable _ -> True
      Unspecified _ _ -> True
      BadJSONResp _ -> True
      -- ...
```

app/Main.hs
```diff
@@ -229,25 +278,45 @@ getHttpManager opts = newManager managerSettings
...
 vaultEnv context = do
-  mountInfo <- requestMountInfo context
-  secrets <- mapExceptT (fmap $ first SecretFileError) $ readSecretList secretFile
-  secretEnv <- requestSecrets context mountInfo secrets
-  checkNoDuplicates (buildEnv secretEnv)
+  mountInfo <- doWithRetries retryPolicy getMountInfo
+  case mountInfo of
+    Left vaultError -> pure $ Left vaultError
+    Right mountInfo' -> do
+      secrets <- readSecretsFile secretFile
```
duijf: Another place you can use `LambdaCase`:

```haskell
Right mountInfo -> readSecretsFile secretFile >>= \case
  Left sfError -> ...
  Right -> ...
```

app/Main.hs
```diff
@@ -229,25 +278,45 @@ getHttpManager opts = newManager managerSettings
...
+--
+-- Signals failure through a value of type VaultError.
+vaultEnv :: Context -> IO (Either VaultError [EnvVar])
 vaultEnv context = do
-  mountInfo <- requestMountInfo context
-  secrets <- mapExceptT (fmap $ first SecretFileError) $ readSecretList secretFile
-  secretEnv <- requestSecrets context mountInfo secrets
-  checkNoDuplicates (buildEnv secretEnv)
+  mountInfo <- doWithRetries retryPolicy getMountInfo
+  case mountInfo of
```
duijf: Opportunity for `LambdaCase`:

```haskell
doWithRetries retryPolicy getMountInfo >>= \case
  Left vaultError -> ...
  Right mountInfo -> ...
```

app/Main.hs
```diff
@@ -229,25 +278,45 @@ getHttpManager opts = newManager managerSettings
...
+    Left vaultError -> pure $ Left vaultError
+    Right mountInfo' -> do
+      secrets <- readSecretsFile secretFile
+      case secrets of
+        Left sfError -> pure $ Left $ SecretFileError sfError
+        Right secrets' -> do
+          secretEnv <- requestSecrets context mountInfo' secrets'
```
duijf: And another one

app/Main.hs
maartenberg: Good one, fixed.

app/Main.hs
maartenberg: Done!

app/Main.hs
maartenberg: Done!

app/Main.hs
maartenberg: Done!

PR #91, opened 2020-03-02
Fix permission issue on .deb package

maartenberg:
We've noticed an issue with the permissions of the file installed from the .deb
package which caused the installed binary and `/etc/secrets.d` to not be owned by
`root` but by the uid of the building user on the build machine.

This PR alters the packaging script to use `fakeroot` to ensure that the built
package includes files owned by `root` instead of some random user, without requiring
building the package as `root`.
It also adds a `postinst` script that will fix the permissions after upgrading from
an affected version of the package.

### DIFF
--- a/package/build_package.sh
+++ b/package/build_package.sh
@@ -18,24 +18,41 @@ export VERSION=$(stack query locals vaultenv-real version | sed -e "s/^'//" -e "
 # The name of the .deb file to create
 PKGNAME="vaultenv-${VERSION}"
 
-# Recreate the file system layout as it should be on the target machine.
-mkdir -p "$PKGNAME/DEBIAN"
-mkdir -p "$PKGNAME/usr/bin"
-mkdir -p "$PKGNAME/etc/secrets.d"
-
 cd ..
 VAULTENV_NIX_PATH=$($(nix-build --no-link -A full-build-script nix/vaultenv-static.nix) | tail -n1)
 cd -
 
-cp "${VAULTENV_NIX_PATH}/bin/vaultenv" "$PKGNAME/usr/bin/"
-cp "${VAULTENV_NIX_PATH}/bin/vaultenv" "vaultenv-${VERSION}-linux-musl"
+cp --no-preserve=mode,ownership "${VAULTENV_NIX_PATH}/bin/vaultenv" "vaultenv-${VERSION}-linux-musl"
 
-# Write the package metadata file, substituting environment variables in the
-# template file.
-cat deb_control | envsubst > "$PKGNAME/DEBIAN/control"
 
 # Build the Debian package
-dpkg-deb --build "$PKGNAME"
+# We need to use `fakeroot` here to have the package files be correctly owned
+# by `root` without being `root` ourselves.
+fakeroot -- bash <<-EOFAKEROOT
+  # Re-enable safe scripting options since this is a new shell
+  set -euf -o pipefail
+
+  # Ensure created directories are mode 0755 (default is 0775)
+  umask 022
+
+  # Recreate the file system layout as it should be on the target machine.
+  mkdir -p "$PKGNAME/DEBIAN"
+  mkdir -p "$PKGNAME/usr/bin"
+  mkdir -p "$PKGNAME/etc/secrets.d"
+
+  # Write the package metadata file, substituting environment variables in the
+  # template file.
+  cat deb_control | envsubst > "$PKGNAME/DEBIAN/control"
+
+  cp deb_postinst "$PKGNAME/DEBIAN/postinst"
+  chmod 0755 "$PKGNAME/DEBIAN/postinst"
+
+  # Copy the built binary from the Nix store to the target directory
+  cp --no-preserve=mode,ownership "${VAULTENV_NIX_PATH}/bin/vaultenv" "$PKGNAME/usr/bin/"
+  chown root:root "$PKGNAME/usr/bin/vaultenv"
+  chmod 0755 "$PKGNAME/usr/bin/vaultenv"
+  dpkg-deb --build "$PKGNAME"
+EOFAKEROOT
 
 # Clean up temporary files
 rm -fr "$PKGNAME"
--- a/package/build_package.sh
+++ b/package/build_package.sh
@@ -13,7 +13,8 @@ set -euf -o pipefail
 
 # Get the package version from Stack, eliminate the single quotes.
 # Exported, because it is used with envsubst to write the control file.
-export VERSION=$(stack query locals vaultenv-real version | sed -e "s/^'//" -e "s/'$//")
+VERSION=$(stack query locals vaultenv-real version | sed -e "s/^'//" -e "s/'$//")
+export VERSION
 
 # The name of the .deb file to create
 PKGNAME="vaultenv-${VERSION}"
--- a/package/build_package.sh
+++ b/package/build_package.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2091,SC2103
 
 # This script builds a .deb package from the static binary build by Nix.
 #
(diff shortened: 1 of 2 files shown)


package/deb_postinst
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
duijf: Now that we're here: can you run shellcheck on this file and on `build_package.sh`?

package/deb_postinst
```diff
@@ -0,0 +1,23 @@
...
+# When installing the package, this script is called with the first argument
+# "configure". If the package is being upgraded from an earlier version, this version is
+# passed as the second argument.
+
+if [ "$1" = 'configure' ] && [ -n "$2" ]; then
+    # If upgrading from an earlier version than 0.13.1, the permissions on
+    # /etc/secrets.d and /usr/bin/vaultenv will be messed up.
+    # We fix this here:
+
+    if dpkg --compare-versions "$2" le 0.13.0; then
```
duijf: Aren't the `chown` and `chmod` commands idempotent? Can we omit the version check (less code to verify the correctness of)?

maartenberg:
> How did you test the changes here?

I've tested the changes by
 - Bumping the version number locally to v0.13.1 and rebuilding the package with the
   new script
 - Installing the package locally and verifying that the ownership and mode of the
   existing files changed as intended
 - Comparing the newly-built package with the published 0.13.0 build using
   `diffoscope --exclude ./usr/bin/vaultenv vaultenv-0.13.0.deb
   vaultenv-0.13.1.deb`,
   which resulted in the following diff:

   ```
   --- /tmp/vaultenv-0.13.0.deb
   +++ vaultenv-0.13.1.deb
   ├── file list
   │ @@ -1,3 +1,3 @@
   │ --rw-r--r--   0        0        0        4 2020-02-19 13:44:42.000000 debian-binary
   │ --rw-r--r--   0        0        0      360 2020-02-19 13:44:42.000000 control.tar.xz
   │ --rw-r--r--   0        0        0  1216948 2020-02-19 13:44:42.000000 data.tar.xz
   │ +-rw-r--r--   0        0        0        4 2020-03-03 13:03:46.000000 debian-binary
   │ +-rw-r--r--   0        0        0      740 2020-03-03 13:03:46.000000 control.tar.xz
   │ +-rw-r--r--   0        0        0  1218560 2020-03-03 13:03:46.000000 data.tar.xz
   ├── control.tar.xz
   │ ├── control.tar
   │ │ ├── file list
   │ │ │ @@ -1,2 +1,3 @@
   │ │ │ -drwxr-xr-x   0 root         (0) root         (0)        0 2020-02-19 13:44:42.000000 ./
   │ │ │ --rw-r--r--   0 root         (0) root         (0)      191 2020-02-19 13:44:42.000000 ./control
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./
   │ │ │ +-rw-r--r--   0 root         (0) root         (0)      191 2020-03-03 13:03:46.000000 ./control
   │ │ │ +-rwxr-xr-x   0 root         (0) root         (0)      764 2020-03-03 13:03:46.000000 ./postinst
   │ │ ├── ./control
   │ │ │ @@ -1,6 +1,6 @@
   │ │ │  Package: vaultenv
   │ │ │ -Version: 0.13.0
   │ │ │ +Version: 0.13.1
   │ │ │  Priority: optional
   │ │ │  Architecture: amd64
   │ │ │  Maintainer: Laurens Duijvesteijn <laurens@channable.com>
   │ │ │  Description: Run processes with secrets from HashiCorp Vault
   ├── data.tar.xz
   │ ├── data.tar
   │ │ ├── file list
   │ │ │ @@ -1,6 +1,6 @@
   │ │ │ -drwxrwxr-x   0 maarten   (1000) maarten   (1000)        0 2020-02-19 13:43:24.000000 ./
   │ │ │ -drwxrwxr-x   0 maarten   (1000) maarten   (1000)        0 2020-02-19 13:43:24.000000 ./etc/
   │ │ │ -drwxrwxr-x   0 maarten   (1000) maarten   (1000)        0 2020-02-19 13:43:24.000000 ./etc/secrets.d/
   │ │ │ -drwxrwxr-x   0 maarten   (1000) maarten   (1000)        0 2020-02-19 13:43:24.000000 ./usr/
   │ │ │ -drwxrwxr-x   0 maarten   (1000) maarten   (1000)        0 2020-02-19 13:44:42.000000 ./usr/bin/
   │ │ │ --r-xr-xr-x   0 maarten   (1000) maarten   (1000)  5314400 2020-02-19 13:44:42.000000 ./usr/bin/vaultenv
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./etc/
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./etc/secrets.d/
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./usr/
   │ │ │ +drwxr-xr-x   0 root         (0) root         (0)        0 2020-03-03 13:03:46.000000 ./usr/bin/
   │ │ │ +-rwxr-xr-x   0 root         (0) root         (0)  5314432 2020-03-03 13:03:46.000000 ./usr/bin/vaultenv
   ```

   This shows the desired changes: the binary and directories are owned by `root`
   instead of uid 1000, and the group-writable bit is no longer set on the installed
   files (Debian convention).

package/deb_postinst
maartenberg: `deb_postinst` was already Shellcheck-clean, `build_package.sh` is now too.

I've silenced two errors:

```
In build_package.sh line 23:
VAULTENV_NIX_PATH=$($(nix-build --no-link -A full-build-script nix/vaultenv-static.nix) | tail -n1)
                    ^-- SC2091: Remove surrounding $() to avoid executing output.


In build_package.sh line 24:
cd -
^-- SC2103: Use a ( subshell ) to avoid having to cd back.
```

This is because the first error is intended (we want to execute here), and the second
solution is not possible because we assign to `VAULTENV_NIX_PATH`.

package/deb_postinst
maartenberg: It should be, I've removed the `if` now.

package/build_package.sh
```diff
@@ -13,29 +14,44 @@ set -euf -o pipefail
...
 
-# Write the package metadata file, substituting environment variables in the
-# template file.
-cat deb_control | envsubst > "$PKGNAME/DEBIAN/control"
 
 # Build the Debian package
-dpkg-deb --build "$PKGNAME"
+# We need to use `fakeroot` here to have the package files be correctly owned
+# by `root` without being `root` ourselves.
+fakeroot -- bash <<-EOFAKEROOT
```
ruuda: Why not wrap the entire `build_package.sh` in `fakeroot`?

maartenberg: I didn't do this because `fakeroot`'s documentation discourages running builds under `fakeroot`:

> Also, note that it's best not to do the building of the binaries themselves under
> fakeroot. Especially configure and friends don't like it when the system suddenly
> behaves differently from what they expect. (or, they randomly unset some
> environment  variables, some of which fakeroot needs).

maartenberg:
> As a final test/precaution: can you also test installing this package together with one of our internal ones which writes to `/etc/secrets.d`?

Tested, works. Rebasing and merging.

package/build_package.sh
ruuda: Ah, I see that `build_package.sh` here also triggers Nix to build the binary, if it wasn't built before. I thought we triggered the build elsewhere and just copied the binary in this script.

PR #95, opened 2020-06-15
Allow VAULT_ADDR addresses without ports

### LINKED ISSUE #94: VAULT_ADDR specification doesn't need to require port number
Current implementation of `vaultenv` diverges from the Vault itself and always requires a port number to be specified in the VAULT_ADDR env variable.

It is a bit misleading because it uses the same variable as many other clients (including the Hashicorp client) but handles it differently.

It would be (a) easier (b) more consistent with other Vault tooling to default the port number to 80 in case of `http://` and 443 in case of `https://`.

Example:

```
$ VAULT_TOKEN=foobar
$ VAULT_ADDR=https://vault.mycompany.com
$ vaultenv --secrets-file vault.secrets python myscript.py
[ERROR] The port //vault.mycompany.com is not a valid int value
```

FPtje:
Fixes #94 

This implementation expands the String operations to get the job done. This works, but it feels rather fragile. Chances are, there are other valid URLs that are not accepted by the code. Perhaps in time it is better to use [parseRequest](https://hackage.haskell.org/package/http-client-0.7.0/docs/Network-HTTP-Client.html#v:parseRequest) or [parseURI](https://hackage.haskell.org/package/network-uri-2.6.3.0/docs/Network-URI.html#v:parseURI).

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -166,6 +166,7 @@ data OptionsError
   | UnknownScheme URI -- ^ The scheme of the address is invalid, e.g. ftp://
   | NonNumericPort String -- ^ The port of the address is not an valid integer
   | HostPortSchemeAddrMismatch String (Maybe Bool) (Maybe String) (Maybe Int) URI
+  | URIPathNotEmpty String -- ^ Disallow using non-empty path at the end of a URL
       -- ^ The source, useTls, host, port and scheme do not match the provided address
   deriving (Eq)
 
@@ -185,6 +186,7 @@ instance Show OptionsError where
       "Hint: This can happen when you provide a `addr` which ",
       "conflicts with `port`, `host`, or `[no-]connect-tls` in either the environment variables or the CLI.\n"
     ]
+  show (URIPathNotEmpty s) = "The address contains a non-empty path \"" ++ s ++ "\". This is not supported."
 
 -- | Validates for a set of options that any provided addr is valid and that either the
 -- scheme, host and port or that any given addr matches the other provided information.
@@ -288,6 +290,9 @@ splitAddress addr = do
   -- exactly right when parsing URIs
   -- See https://github.com/haskell/network-uri/issues/19
   uriAuth <- maybe (Left $ URIParseError addr) Right $ uriAuthority addr
+  when (uriPath addr `notElem` ["", "/"]) $
+    Left $ URIPathNotEmpty $ uriPath addr
+
   scheme <- case uriScheme addr of
     "https:" -> Right HTTPS
     "http:" -> Right HTTP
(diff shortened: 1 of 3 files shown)


test/ConfigSpec.hs
```diff
@@ -45,6 +45,13 @@ spec =
           addr = "http://" ++ host ++ ":" ++ port
         in ':' `notElem` port ==> splitAddress addr `shouldBe` (Just "http://", host, port)
       )
+    it "should parse addr without explicit port" $
+      property (\host ->
+        let
+          addr :: String
+          addr = "http://" ++ host
+        in ':' `notElem` host ==> splitAddress addr `shouldBe` (Just "http://", host, "80")
+      )
```
joris-burgers, line 54: Could you add also tests for the other situations that might occur here, namely `https://` and any other string, for example `ftp://`. You can copy the test or you can create a `newtype` wrapper and give an arbitrary instance for that wrapper, like 
```haskell
it "should parse addr without explicit port" $
      property (\protocol host ->
        let
          addr :: String
          addr = unProtocol protocol ++ "://" ++ host
        in ':' `notElem` host ==> splitAddress addr `shouldBe` (Just (unProtocol protocol ++ "://"), host, getPort protocol)
      )
```

but I leave it up to you if you want that.

FPtje: I've added a test for `https://`. Other schemas appear to be rejected, as described by the `should reject invalid schemes` test.

test/ConfigSpec.hs
```diff
@@ -45,6 +45,18 @@ spec =
           addr = "http://" ++ host ++ ":" ++ port
         in ':' `notElem` port ==> splitAddress addr `shouldBe` (Just "http://", host, port)
       )
+    it "should parse addr without explicit port" $
```
duijf, line 48: Thanks for adding tests!

FPtje:
Just refactored the code to use `parseURI`. The `splitAddress` function can now throw some meaningful errors. That's because during the splitting is the point where many issues are found.

There is some quirk with `parseURI`, though, where URIs without a scheme get parsed, but end up having unexpected schemes and `uriAuthority` values:
```
> uriScheme <$> parseURI "localhost:80"
Just "localhost:"
-- On these "successful but pretty wrong" parses uriAuthority is Nothing.
> uriAuthority <$> parseURI "localhost:80"
Just Nothing
-- Just a hostname without scheme or port fails to parse
> parseURI "localhost"
Nothing
```
There's an issue about this here: https://github.com/haskell/network-uri/issues/19

src/Config.hs
```diff
@@ -404,6 +398,13 @@ parseEnvOptions envVars
         Just "error" -> Just Error
         Nothing -> Nothing
         _ -> err key
+    -- | Look up and parse VAULD_ADDR
```
duijf, line 401:
```suggestion
    -- | Look up and parse VAULT_ADDR
```

FPtje: Whoops, thanks!

FPtje:
An error is now thrown when a non-empty path is given in the address. One could imagine using paths for situations where the vault sits behind e.g. an nginx server (`https://example.com/vault/`). The original vault tool seems to support this, but I'm not sure whether this feature would actually be in demand.

PR #97, opened 2021-01-14
Make vaultenv more friendly for use with Vault Agent

bertptrs:
Vault Agent is a kind of local Vault proxy that can take care of authentication and caching, but it's better described by [its own documentation](https://www.vaultproject.io/docs/agent).

This PR makes Vaultenv easier to use with Vault agent, with two changes:

- The `Token` parameter is now optional, as it is not required when the Vault Agent provides auto authentication. If the parameter is not set, the relevant header is not added to the request. If the token turned out to be required after all, the Vault server will respond with a `BadRequest` informing you of that fact so it still fails fast.
- The `x-vault-request: true` header is added to every request. If Vault's SSRF protection is enabled, this header is required for requests to go through. Otherwise, it is silently ignored, just like any other optional header.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -361,6 +361,7 @@ requestMountInfo context =
               (cHttpManager context)
         $ setRequestHeader "x-vault-token"
               [SBS.pack (getOptionsValue oVaultToken cliOptions)]
+        $ setRequestHeader "x-vault-request" [SBS.pack "true"]
         $ setRequestPath
               (SBS.pack "/v1/sys/mounts")
         $ setRequestPort
@@ -393,6 +394,7 @@ requestSecret context secretPath =
         = setRequestManager
             (cHttpManager context)
         $ setRequestHeader "x-vault-token" [SBS.pack (getOptionsValue oVaultToken cliOptions)]
+        $ setRequestHeader "x-vault-request" [SBS.pack "true"]
         $ setRequestPath    (SBS.pack secretPath)
         $ setRequestPort    (getOptionsValue oVaultPort cliOptions)
         $ setRequestHost    (SBS.pack (getOptionsValue oVaultHost cliOptions))
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -351,6 +351,19 @@ runCommand options env =
     -- replaces the current process with `command`. It does not return.
     executeFile command searchPath args env'
 
+
+-- | Add Vault authentication token to a request, if set in the options. If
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -359,8 +372,8 @@ requestMountInfo context =
     request
         = setRequestManager
               (cHttpManager context)
-        $ setRequestHeader "x-vault-token"
-              [SBS.pack (getOptionsValue oVaultToken cliOptions)]
+        $ addVaultToken cliOptions
+        $ setRequestHeader "x-vault-request" [SBS.pack "true"]
```
ruuda, line 376: The `SBS.pack` is not needed here, because [`BytesString` implements `IsString`](https://www.stackage.org/haddock/lts-16.29/bytestring-0.10.10.1/Data-ByteString.html#t:ByteString), and because `OverloadedStrings` is enabled at the top of this file. With `OverloadedStrings` enabled, string literals can have any type `IsString a => a`, not just `String`, so `"true"` can be a `ByteString` literal.

app/Main.hs
```diff
@@ -392,7 +405,8 @@ requestSecret context secretPath =
     request
         = setRequestManager
             (cHttpManager context)
-        $ setRequestHeader "x-vault-token" [SBS.pack (getOptionsValue oVaultToken cliOptions)]
+        $ addVaultToken     cliOptions
+        $ setRequestHeader "x-vault-request" [SBS.pack "true"]
```
ruuda, line 409: The `pack` is not needed here either.

app/Main.hs
```diff
@@ -351,6 +351,19 @@ runCommand options env =
...
 
+
+-- | Add Vault authentication token to a request, if set in the options. If
+-- not, the request is returned unmodified.
+addVaultToken :: Options Validated Completed -> Request -> Request
+addVaultToken options request =
+  let
+    optToken = oVaultToken options
+  in
+    case optToken of
```
ruuda, line 362: I don’t mind this `let`, but in case you did not realize, you can also just write

```haskell
case oVaultToken options of
  ...
```

ruuda (commented):
The logic looks good! I have two small style suggestions.

app/Main.hs
bertptrs: Thanks! I think I like the `let` expression here, so I think I'll leave it as-is.

ruuda (approved):
LGTM!

PR #98, opened 2021-01-22
Add nix derivation

joris-burgers:
Adds a specialized nix derivation for vaultenv.

Vaultenv can now be included in other projects by building `./nix/release.nix`. The nix dependencies are now also managed using [`niv`](https://github.com/nmattia/niv).

This PR also bumps the GHC version to 8.8.4.


nix/release.nix
```diff
@@ -0,0 +1,7 @@
+# This file wires up the different vaultenv releases that are available.
```
FPtje, line 1: There is only one release available, does this comment still apply?

nix/release.nix
```diff
@@ -0,0 +1,7 @@
...
+rec {
+  pkgs = import ./nixpkgs-pinned.nix {};
```
FPtje, line 3: This doesn't seem like it should be member of the set of released packages. Perhaps it fits better in a let-binding.

stack.yaml.lock
```diff
@@ -4,9 +4,4 @@
 #   https://docs.haskellstack.org/en/stable/lock_files
 
 packages: []
-snapshots:
-- completed:
-    size: 523700
-    url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/14/7.yaml
-    sha256: 8e3f3c894be74d71fa4bf085e0a8baae7e4d7622d07ea31a52736b80f8b9bb1a
-  original: lts-14.7
+snapshots: []
```
FPtje, line 7: I think this file can be removed entirely when using nix 👍

vaultenv.nix
```diff
@@ -0,0 +1,76 @@
+{ haskellPackages
+, mkDerivation
+, lib
+, pkgs
+, glibcLocales
```
FPtje, line 5: Are you sure the `glibcLocales` are needed for this derivation?

FPtje (changes requested):
Thanks for creating better nix packaging and upgrading ghc! Some comments.

nix/release.nix
joris-burgers: Well spotted, no. This PR doesn't use building static binaries, so we only have a single release.

nix/release.nix
joris-burgers: I moved it to a let binding.

stack.yaml.lock
joris-burgers: YES WE CAN!

vaultenv.nix
joris-burgers: Not sure, but I didn't test whether `vaultenv` still suffers from the `commitAndReleaseBuffer: invalid argument (invalid character)` error message without it, that's why I included this.

FPtje (approved):
LGTM! :+1:

PR #100, opened 2021-02-01
Fix `build_package.sh`

joris-burgers:
Fixes `build_package.sh`, that broke after https://github.com/channable/vaultenv/pull/98

This PR:
- Reverts the GHC version back to 8.6.5
- Uses the correct GHC version when using stack
- Allows building of static packages again, by using a separate `.yaml` file.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -10,5 +10,17 @@
         "type": "tarball",
         "url": "https://github.com/NixOS/nixpkgs/archive/65f4c39a40e6ed4343dd94017c3ed81f416cd3b4.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "static-haskell-nix": {
+        "branch": "master",
+        "description": "easily build most Haskell programs into fully static Linux executables",
+        "homepage": "",
+        "owner": "nh2",
+        "repo": "static-haskell-nix",
+        "rev": "ff7715e0e13fb3f615e64a8d8c2e43faa4429b0f",
+        "sha256": "17ir87i7sah9nixvh25qhzh19bqv3vgnfg4nfy4wv631q4gfj7fb",
+        "type": "tarball",
+        "url": "https://github.com/nh2/static-haskell-nix/archive/ff7715e0e13fb3f615e64a8d8c2e43faa4429b0f.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     }
 }
--- a/nix/vaultenv-static.nix
+++ b/nix/vaultenv-static.nix
@@ -1,0 +1,90 @@
+# This is the default build script taken from static-haskell-nix.
+# We've adjusted some things and removed some instructions that
+# we didn't have a usecase for.
+#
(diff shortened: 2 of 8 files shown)


nix/sources.json
```diff
@@ -10,5 +10,17 @@
...
+        "description": "easily build most Haskell programs into fully static Linux executables",
+        "homepage": "",
+        "owner": "nh2",
+        "repo": "static-haskell-nix",
+        "rev": "ff7715e0e13fb3f615e64a8d8c2e43faa4429b0f",
+        "sha256": "17ir87i7sah9nixvh25qhzh19bqv3vgnfg4nfy4wv631q4gfj7fb",
+        "type": "tarball",
+        "url": "https://github.com/nh2/static-haskell-nix/archive/ff7715e0e13fb3f615e64a8d8c2e43faa4429b0f.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+}
```
FPtje, line 25: Nit: This `}` is on the wrong indentation level. Running `niv update static-haskell-nix` will probably cause `niv` to reformat the file :+1:

nix/static-haskell-nix.nix
```diff
@@ -0,0 +1,9 @@
+# Pin static-haskell-nix version.
+let
+  static-haskell-nix-rev = "ff7715e0e13fb3f615e64a8d8c2e43faa4429b0f";
```
FPtje, line 3: This file can be replaced with the `static-haskell-nix` attribute produced by importing `sources.nix`.

FPtje (changes requested):
Nice work! Some minor comments

nix/sources.json
joris-burgers: Well spotted, reformatted it.

nix/static-haskell-nix.nix
joris-burgers: Yes, using the `sources.json` is a good idea. It is updated.

nix/vaultenv-static.nix
```diff
@@ -0,0 +1,89 @@
...
+  # vaultenv also being included in `nixpgks`, which leads to conflicts.
+  cabal-package-name = "vaultenv-real";
+
+  # This has to match the compiler used in the Stackage snapshot.
+  # Update this when the Stackage snapshot changes the version of
+  # GHC it uses.
+  compiler = "ghc865";
+
+  # Pin versions of static-haskell-nix and nixpkgs.
+  static-haskell-nix = import ./static-haskell-nix.nix;
```
FPtje, line 41: Perhaps instead of importing the `static-haskell-nix.nix` file:

```nix
sources = import ./sources.nix;
static-haskell-nix = sources.static-haskell-nix;
```

That way we can delete the `static-haskell-nix.nix` file, which right now only contains 4 lines of code :+1:

FPtje (changes requested):

nix/vaultenv-static.nix
joris-burgers: Even cleaner, good suggestion!

FPtje (approved):
LGTM! :+1:

PR #105, opened 2021-05-27
Use nix derivation in CI

sarahhodne:
Currently CI seems to be failing due to the stack project not working without Nix set up. This pulls in Nix in CI and uses nix to run the build and test commands. Since Nix will handle downloading Stack, we no longer need to install it manually either.

I removed `--copy-bins` as well, since it doesn't seem to be necessary and seems to error for me locally when run using Nix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: true
-language: generic
+language: nix
 
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


bertptrs:
Hi Sarah, thank you for your contribution. As you're probably aware, the Vaultenv source code is licensed as BSD-3-Clause. Should we merge your changes, they will be included under the same terms. Do you accept this?

sarahhodne:
> Hi Sarah, thank you for your contribution. As you're probably aware, the Vaultenv source code is licensed as BSD-3-Clause. Should we merge your changes, they will be included under the same terms. Do you accept this?

Yes, I do.

bertptrs (approved):

Issue #106, opened 2021-06-10
Implement connection reuse


bertptrs:
Vaultenv currently opens a new connection for every request it sends. This causes an unnecessarily heavy load on the receiving Vault server as every request requires a new TCP socket and potentially a TLS handshake. We can avoid this by reusing connections using `Connection: keepalive`.

As far as I can tell, the `Network.HTTP.Client` has support for this and we just don't use it yet. The requests are also missing the required `Connection` header.

fatho:
The single shared HTTP manager should already ensure that connections are reused where possible, as connection reuse is a property of the `Manager` it seems.

But the reason why it's not happening in practice might be that we send all of the requests in parallel:

https://github.com/channable/vaultenv/blob/fe8ccad1e4f50710e53749f901ca3812cd6bd61f/app/Main.hs#L430-L431

Apparently, `mapConcurrently` performs *all* actions concurrently, so maybe just limiting that to a maximum of `n` concurrent requests (where n is significantly lower than the total amount of secrets to fetch) would help.

bertptrs:
That's part of the mystery and it also helps to explain why loading certain secrets file is such a heavy load. However, if I intercept the request headers of a HTTP request from `vaultenv`, I still don't see the correct `Connection` header:

```bash
# In one terminal
nc -l 9001

# In another:
VAULT_ADDR=http://localhost:9001 vaultenv true

# Shown in the nc terminal
GET /v1/sys/mounts HTTP/1.1
Host: localhost:9001
Accept-Encoding: gzip
x-vault-request: true
x-vault-token: REDACTED
```

If connection reuse was enabled, I'd expect to see a `Connection: keep-alive` header.

fatho:
> If connection reuse was enabled, I'd expect to see a `Connection: keep-alive` header.

According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection, "having a persistent connection is the default on HTTP/1.1 requests", so I'd assume that's not a requirement.

PR #107, opened 2021-06-29
Limit parallelism, enabling more connection reuse

fatho:
When there are a lot of secrets to be fetched, imposing a limit makes it
possible to reuse more connections to the vault server. This reduces the
load on the vault server because expensive TLS handshakes can be
avoided.

Closes https://github.com/channable/vaultenv/issues/106.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,15 +99,17 @@ of `tutorial.sh`.
 ## Usage
 
 ```
-vaultenv 0.11.0 - run programs with secrets from HashiCorp Vault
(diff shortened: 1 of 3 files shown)


joris-burgers (commented):
Looks like a clean implementation! I have one request for a comment and can you maybe update the docs in https://github.com/channable/vaultenv/blob/master/README.md#L295?

joris-burgers (approved):
LGTM!

PR #114, opened 2021-10-07
Fix integration tests

ruuda:
I was unable to run the integration tests, for a few reasons:

 * `vault secrets disable` fails with
   
   ```
    Error disabling secrets engine at secret/: Delete "http://localhost:8200/v1/sys/mounts/secret": dial tcp: lookup localhost: no such host
    ```
   
   When I change `localhost` to `127.0.0.1` this error goes away.

* The `VAULT_*` variables set in `integration_test.sh` do not propagate into Vaultenv when Stack uses Nix; it clears the environment, unless we pass `--no-nix-pure`.

* Python 3.7 was hard-coded. Add Python to the Nix development environment instead.

* In the Python integration test, `subprocess.run` with `env` set lost the original `PATH`, and when `vault` is taken from the Nix store, we need to preserve that `PATH`.

With these changes, I can run the integration tests, and they all pass.


maartenberg:
> Semaphore doesn’t seem to run on pull requests that I open

Hi! I've enabled "approval mode" for Semaphore now, so the CI should run if we comment the following on a PR:

/sem-approve

PR #115, opened 2021-10-13
Enable line buffering

ruuda:
When stdout and stderr are not a terminal, like when running under Systemd or Kubernetes, by default the runtime will use block buffering. This prevents log messages from showing up when they are logged, and it breaks interleaving of stdout and stderr. Fix this by explicitly selecting line buffering.


OpsBotPrime:
Pull request approved for merge by @maartenberg, waiting for rebase behind one pull request.

PR #116, opened 2021-11-15
Add support for Kubernetes authentication

ruuda:
## Background

Currently, Vaultenv requires you to have a Vault token. When used in Kubernetes, typically this Vault token is not available directly. Instead, Kubernetes mounts a file into `/var/run/secrets` that contains a Kubernetes service account token, and you can then use this token with [the Vault Kubernetes authentication endpoint](https://www.vaultproject.io/docs/auth/kubernetes) to obtain the Vault token. This authentication method also requires specifying a _role_.

## Changes

Add a new option, `--kubernetes-role`, which is an alternative to `--vault-token`. If `--kubernetes-role` is supplied, Vaultenv will read the Kubernetes token from `/var/run/secrets/kubernetes.io/serviceaccount/token`, authenticate to Vault with that token and the given role, and that produces a token which Vaultenv will use for subsequent requests.

## Testing

This feature is inherently complex: it requires a Kubernetes cluster, which is already a very complex piece of software that consists of multiple moving parts. On top of that it requires a Vault server, and then Vault needs to be configured to use Kubernetes. I wrote a Python script to test Vaultenv with Kubernetes authentication, and eventually I got it working on my machine with Minikube, but it is not fully automated, and likely the test will not work out of the box in other environments. Making this run on CI would be quite an effort and this is not my goal right now.

In addition to this semi-manual testing, I will be testing this against a production Kubernetes cluster before marking this pull request as ready for review.

## Alternatives

Apparently, newer versions of Kubernetes can act as an OIDC provider, and [Vault supports OIDC authentication as well](https://www.vaultproject.io/docs/auth/jwt). OIDC authentication is useful outside of Kubernetes. So maybe this approach of using Kubernetes authentication in Vault is already obsolete. Still, it’s a good way to explore what alternative authentication methods can look like, and I think the work in this PR will make it easier to add OIDC support in the future.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -427,6 +427,15 @@ readKubernetesJwt =
       , Handler $ \(_ :: IOError) -> pure $ Left KubernetesJwtFailedRead
       ]
 
+-- | The "client_token" field from an /auth/kubernetes/login response.
+newtype ClientToken = ClientToken Text
+
+instance FromJSON ClientToken where
+  parseJSON = Aeson.withObject "AuthResponse" $ \obj -> do
+    auth <- obj .: "auth"
+    clientToken <- auth .: "client_token"
+    pure $ ClientToken clientToken
+
 -- | Authenticate using Kubernetes auth, see https://www.vaultproject.io/docs/auth/kubernetes.
 requestKubernetesVaultToken :: Context -> Text -> IO (Either VaultError Text)
 requestKubernetesVaultToken context role = do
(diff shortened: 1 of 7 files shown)


Riscky (commented):
Nice work!
I don't have any specific comments, but I do have a general question: since we want to add more authentication methods later (Github auth to be specific), can you move the Kubernetes specific logic in `Main.hs` to a separate file?

PR #123, opened 2022-06-13
Fix the static build

Riscky:
The static build has been broken since 4cd7431d4c51d01b1e93dbd613de2668fb922489, due to a package upgrade.
This PR fixes the static build (big thanks to @FPtje).
The new solution is a lot simpler, so some old cruft has been removed.

This also adds the build script to CI so our master will be green in the future.
The machine type of the CI runner has also been changed, as we need more disk apparently.


.semaphore/semaphore.yml
```diff
@@ -33,11 +37,22 @@ blocks:
             - "sh ./install-nix-2.3.15 --no-daemon"
             - ". $HOME/.nix-profile/etc/profile.d/nix.sh"
 
+            # Install Cachix and use the Channable cache
+            - "nix-env -iA nixpkgs.cachix"
+            - "cachix use channable-public"
+
+            # Build the devenv and the static package and upload to Cachix
+            - "nix-build --no-out-link default.nix nix/release.nix | cachix push channable-public"
```
ReinierMaas, line 45: What happens here when `nix-build` fails does the semaphore job still succeed? We had such an issue recently in Hoff. That was fixed by splitting those combined steps into separate steps, see https://github.com/channable/hoff/pull/120/files

nix/release.nix
```diff
@@ -3,4 +3,19 @@ let
...
   # Normal cabal build where Nix handles dependencies.
   vaultenv = pkgs.haskellPackages.callPackage ../vaultenv.nix {};
+
+  # Static package build
+  vaultenvStatic = (pkgs.pkgsStatic.haskellPackages.callPackage ../vaultenv.nix {
+    # Use non-static version of glibc locales, as the static version fails to build.
+    glibcLocales = pkgs.glibcLocales;
+  }).overrideAttrs (prev: {
+    # Disable the check phase, as hspec-discover will not run.
+    # TODO: investigate what's happening with hspec-discover.
```
ReinierMaas, line 13: Do you still intend to debug the `TODO`s? Could you make issues otherwise?

nix/haskell-overlay.nix
```diff
@@ -1,3 +1,2 @@
 self: super: {
-  hashable = super.hashable_1_4_0_2;
```
ReinierMaas, line 2: Nice that the overlay was resolved with the nix pin update!

vaultenv.nix
```diff
@@ -53,20 +53,20 @@ mkDerivation {
...
-    # a dependency on GHC and other build tools. Those are not necessary for running the
-    # executable, so we get rid of that.
-    rm -rf $out/lib
-  '';
+  # postInstall = ''
+  #   # By default, a Haskell derivation contains extra stuff that causes the output to retain
+  #   # a dependency on GHC and other build tools. Those are not necessary for running the
+  #   # executable, so we get rid of that.
+  #   rm -rf $out/lib
+  # '';
```
ReinierMaas, line 61: Was this intended to be commented out? It sounds important to keep the shipped derivation small.

ReinierMaas (changes requested):
A few small nits but on the overarching picture, this is almost done!

.semaphore/semaphore.yml
Riscky: The pipeline would still fail as the next commands also use the same builds and would fail, but this is merely by coincidence.
I'll change the pipe to a redirection.

nix/haskell-overlay.nix
Riscky: I was quite surprised tbh.
Building hashable took forever, so I tried taking it out, and it worked.

vaultenv.nix
Riscky: I have no clue why @FPtje commented this out, it seems to work fine when not commented out.

nix/release.nix
Riscky: Both TODO's don't seem overly important.
The first because the tests are run elsewhere anywhere.
The second because it is just a slightly ugly, but still functional, solution.

Riscky: It doesn't pick up the tests when enabled.
Since the `pkgsStatic.haskellPackages` call is, in Falco's words, 'magic', I think I lack the knowledge to debug this.

ReinierMaas (approved):
LGTM!

Riscky:
@OpsBotPrime merge

OpsBotPrime:
Pull request approved for merge by @Riscky, rebasing now.

Issue #125, opened 2022-06-13
Small reproducibility issue with the 0.15.0 release


ruuda:
Thanks for fixing the static build in #123! I just tried to see if I could reproduce it, so I can safely download the binary from GitHub and confirm the shasum for my downstream use case, instead of having to build it from source. However, I don’t get exactly the same binary as the one on the release page.

### Differences

```
0960ab3f51abb71e4129c5fe0541e3dad4c2ae5428d51b757f2f3cf5c9ff4758  result/bin/vaultenv
30273ff1511baef0ac2eb5606f2c54b4d4ec46965344f56a3b1088ce535d6d90  ~/downloads/vaultenv-0.15.0-linux-musl
```

I am building commit 3b55e1770c244a6bb212cb294ecc301c13c88825, better known as `v0.15.0`, with `nix-build nix/release.nix -A vaultenvStatic`.

Diffoscope shows very few differences (this is good, it means the discrepancy is hopefully easy to resolve), all caused by a single different store path which ends up in various places in the binary.
* `/nix/store/y9p11057zpc5r9cqnfy2sgfgjymwv59m-vaultenv-static-x86_64-unknown-linux-musl-0.15.0` is what’s in the published release.
* `/nix/store/kkb4w0i0qvhj60kdnxpasldq79glnmvg-vaultenv-static-x86_64-unknown-linux-musl-0.15.0` is what I get in my local build.

<details>
<summary>Diffoscope output</summary>

```
--- ~/downloads/vaultenv-0.15.0-linux-musl
+++ result/bin/vaultenv
├── strings --all --bytes=8 {}
│ @@ -931,25 +931,25 @@
│  ath component
│  hitespace
│  vaultenv-real-0.15.0-BVsC6AJuNM75bSImrOaUmP:SecretsFile.IOErr
│  vaultenv-real-0.15.0-BVsC6AJuNM75bSImrOaUmP:SecretsFile.ParseErr
│  vaultenv-real-0.15.0-BVsC6AJuNM75bSImrOaUmP:SecretsFile.V1
│  vaultenv-real-0.15.0-BVsC6AJuNM75bSImrOaUmP:SecretsFile.V2
│  vaultenv-real-0.15.0-BVsC6AJuNM75bSImrOaUmP:SecretsFile.Secret
│ -/nix/store/y9p11057zpc5r9cqnfy2sgfgjymwv59m-vaultenv-static-x86_64-unknown-linux-musl-0.15.0/etc
[...]

PR #126, opened 2022-06-20
Ignore integration test artifacts in Nix build

Riscky:
This removes files that are generated by the integration tests from the Nix `src` attribute.
This makes sure the build is reproducible (see issue #125).

### DIFF
--- a/vaultenv.nix
+++ b/vaultenv.nix
@@ -3,6 +3,7 @@
 , lib
 , pkgs
 , glibcLocales
+, nix-gitignore
 }:
 let
   dependencies = import ./nix/haskell-dependencies.nix haskellPackages;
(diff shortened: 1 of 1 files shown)


vaultenv.nix
```diff
@@ -31,8 +31,8 @@ mkDerivation {
       blacklistedSrc = lib.cleanSourceWith {
         src = whitelistedSrc;
         filter = path: type:
-          # Where we're going we don't need no documentation
-          ! lib.hasSuffix ".md" path;
+          # Where we're going we don't need no documentation (nor caches)
+          ! (lib.hasSuffix ".md" path || lib.hasSuffix ".cache" path || lib.hasSuffix "__pycache__" path || lib.hasSuffix ".tar.gz" path);
```
ReinierMaas, line 35: We should probably also exclude the `.gitignore`d files:

```nix
      # Remove all things we don't need such as python caches that otherwise mess up the build
      # by causing unnecessary rebuilds due to supposedly changed inputs.
      blacklistedSrc = lib.cleanSourceWith {
        src = whitelistedSrc;
        filter =
          let
            gitignore = builtins.readFile ./.gitignore;
            extraFilter = path: type:
              # Prevent python from messing with the rebuilds
              ! (  lib.hasInfix "/__pycache__/" path
                || lib.hasSuffix "/__pycache__" path
              # Where we're going we don't need no documentation
                || lib.hasSuffix ".md" path
                );
          in
            nix-gitignore.gitignoreFilterPure extraFilter gitignore ./.
          ;
      };
```

What tool created the `.cache` suffix files?

Riscky: > We should probably also exclude the .gitignored files:

done

> What tool created the .cache suffix files?

Some step in the integration tests, but I haven't been able to pinpoint which

ReinierMaas (approved):
LGTM! Thanks for incorporating both `.gitignore` files.

Riscky:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 128b11b75955350d155e76d6486b7633796a6ccc, waiting for CI …

PR #127, opened 2022-06-28
Add `VAULT_TOKEN` to command environment

crtschin:
When using Github/kubernetes as the authentication method, we request a token from vault to make subsequent requests. 
This PR additionally adds that token to the child command's environment in the `VAULT_TOKEN` variable.


ReinierMaas (approved):
LGTM! This should indeed resolve the missing token we were seeing.

bertptrs:
Can confirm that this fixes it, thanks! I'll merge this and craft a release.

@OpsBotPrime merge

Issue #99, opened 2021-01-31
vaultenv fails with HandshakeFailed error rejecting the certificate by NameMismatch


sezaru:
Hello,

When vaultenv tries to connect to my vault server using TLS it fails with the following message:

``` sh
vaultenv --addr https://127.0.0.1:8200 --token s.11iJ3aza8NC0UL1TWN2koAW3 --secrets-file /tutorial.secrets /script.sh
[ERROR] ServerUnreachable error: HttpExceptionRequest Request {
  host                 = "127.0.0.1"
  port                 = 8200
  secure               = True
  requestHeaders       = [("x-vault-token","**removed**")]
  path                 = "/v1/sys/mounts"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HandshakeFailed (Error_Protocol ("certificate rejected: [NameMismatch \"127.0.0.1\"]",True,CertificateUnknown))))
```

My vault uses a certificate and key that was signed by my private CA. I will attach them here in the hope it helps resolve the issue (don't worry, these certs, keys, token, etc are all temporary, is not the ones in production).

[certificates_and_keys.zip](https://github.com/channable/vaultenv/files/5899174/certificates_and_keys.zip)

Bellow is the output vault gives me in its log (journalctl -u vault.service):
```
Jan 31 04:05:01 tip-off vault[744]: 2021-01-31T04:05:01.213Z [INFO]  http: TLS handshake error from 127.0.0.1:46152: remote error: tls: unknown certificate
```

Also, using vault cli works without issue.

Any idea of what this could be?

ruuda:
Indeed, this is a limitation of the Haskell TLS client library that Vaultenv uses. It looks like there are two solutions in https://github.com/kubernetes-client/haskell/issues/64:

* Depend on a custom version of hs-certificate where https://github.com/vincenthz/hs-certificate/pull/113 is applied, but looking at that pull request, it is not ready for production use.
* Depend on `http-client-openssl` instead of `http-client-tls`. I think that would be the best way to go. We use `http-client-openssl` in a different (internal) project, and it requires some fiddling to configure, but I think it would be worth it; it would also sidestep the problem that versions built with `integer-simple` rather than `integer-gmp` are extremely slow. On the other hand, it would also make it even harder to build a static binary then.

PR #137, opened 2023-12-20
Use http-client-openssl instead of http-client-tls

jfroche:
Using http-client-openssl that also handle the SSL_CERT_FILE environment variable

refs #99

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -25,7 +25,8 @@ import Network.HTTP.Simple    (HttpException(..), Request, Response,
 import Network.HTTP.Client.OpenSSL (defaultMakeContext, defaultOpenSSLSettings,
                                      opensslManagerSettings, osslSettingsVerifyMode)
 import OpenSSL.Session        (contextSetDefaultVerifyPaths,
-                               VerificationMode (VerifyNone, VerifyPeer))
+                               VerificationMode (VerifyNone, VerifyPeer),
+                               vpFailIfNoPeerCert, vpClientOnce, vpCallback)
 import System.Environment     (getEnvironment)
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -298,16 +301,22 @@ getHttpManager opts = newManager $ applyConfig basicManagerSettings
...
+        { managerConnCount = maxConnections }
+    makeContext = do
+      context <- defaultMakeContext opensslSettings
+      contextSetDefaultVerifyPaths context
+      pure context
+    opensslSettings = defaultOpenSSLSettings
+      { osslSettingsVerifyMode =
+            if not $ getOptionsValue oValidateCerts opts
+            then VerifyNone
+            else VerifyPeer
```
HugoPeters1024, line 315: Could you use record syntax here to make it more clear what these settings do?

package.yaml
```diff
@@ -28,7 +30,7 @@ dependencies:
   - utf8-string
   - optparse-applicative
 
-ghc-options: -Wall -Werror
+ghc-options: -threaded -rtsopts -Wall -Werror
```
HugoPeters1024, line 33: Could you explain why these extra options are necessary?

jfroche: I am not an expert in this openssl library but without the `-threaded` option we face
`ConnectionFailure user error (RTS doesn't support multiple OS threads (use ghc -threaded when linking))` at run time when trying to connect to vault. So it seems that openssl requires a threaded runtime.

`-rtsopts` enable all RTS options processing and can be configured at runtime through command line or environment variable. It enables users to configure the number of threads, the heap size, ... We could remove it until we know that we really need it as it might have security implications.

HugoPeters1024 (approved):
Thank you for taking the time to make a contribution :)

LGTM!

PR #139, opened 2024-02-22
Add support for comments in secrets files

ruuda:
## Motivation

I shot myself in my foot because of [this line in the readme](https://github.com/channable/vaultenv/blob/c61036674d3a385523b38e5b30105148e91f6cfd/README.md?plain=1#L295):

> Also: comments are allowed if they start with `#`.

The comment is about `vaultenv.conf`, not about the secrets file, so it’s entirely my own fault. Still, what happens when you mistakenly think that secrets files support comments, and you feed Vaultenv this file?

```
VERSION 2
MOUNT secret
FOO=app#foo

# Comment
BAR=app#bar
```

Then Vaultenv will parse up to the comment, and silently drop the `BAR=` definition.

## Changes

1. Add a new test for a file that has comments. It fails before changing the parser.
2. Extend the tests; before this they only tested that the files can be parsed or not parsed, but they don’t check what the parsed result looks like.
3. Rewrite the parser to support comments.

There was a Megaparsec parser but I found it difficult to introduce comment support. I think using a parser for this creates more confusion (and danger, as it dropped half my input) than a simple split-on-lines, split-on-`=`, split-on-`#`. So I rewrote it without Megaparsec. It is now 100 lines smaller and supports comments.

This makes some breaking changes: the version line must be `VERSION 2`, other whitespace is no longer allowed. I think that’s acceptable but if you prefer to keep it compatible it’s also not too difficult. Either way if you write `VERSION    2` that will fail to parse and not silently result in an unexpected secrets list. Also it is now a bit more lax about what path names and key names it accepts.

### DIFF
--- a/test/SecretFileSpec.hs
+++ b/test/SecretFileSpec.hs
@@ -4,6 +4,7 @@ import Test.Hspec
 
 import Data.Either (isRight, isLeft)
 
+
 import qualified SecretsFile
 import qualified System.Directory as Dir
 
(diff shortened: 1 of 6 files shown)


test/SecretFileSpec.hs
```diff
@@ -3,21 +3,119 @@ module SecretFileSpec where
...
 import qualified System.Directory as Dir
 
 spec :: SpecWith ()
 spec = do
   describe "SecretFile.readSecretList" $ do
-    it "parses all golden tests succesfully" $ do
-      goldenTestContents <- Dir.listDirectory "test/golden"
-      let goldenTestFiles = map ("test/golden/" <>) goldenTestContents
-      parseResults <- mapM SecretsFile.readSecretsFile goldenTestFiles
-      parseResults `shouldSatisfy` (all isRight)
```
ruuda, line 17: I changed this to have one `it` per file, so that when the tests fail, it only prints the failing file, instead of a big list with and “`all isRight` is false for this big list, good luck finding the offending one!”.

FPtje:
Thanks for the PR @ruuda and thanks for the input @duijf!

The production outage chronicled by @duijf seems to be a good argument to keep the megaparsec library. Along with it being a breaking change could mean that an update of `vaultenv` could start breaking production, causing a parse error on startup.

Perhaps we can take another look at the challenge of implementing the comment parser in Megaparsec. The rest of the PR looks great.

fatho:
> and danger, as it dropped half my input

Are we perhaps missing a [`eof`](https://hackage.haskell.org/package/megaparsec-9.6.1/docs/Text-Megaparsec.html#v:eof) at the end of the parser to make sure it either always consumes the whole document or dies trying?

PR #140, opened 2024-04-08
Nixpkgs update, and crypton-connection upgrade

ErnestKz:
Seems like `crypton-connection` is superseding `connection`, and a large portion of the Haskell ecosystem has done the migration. Having both connection and crypton-connection in a project causes a conflict on the `tls` version.

Links:
- https://hackage.haskell.org/package/crypton-connection-0.3.1
- https://hackage.haskell.org/package/connection

PR #143, opened 2024-07-23
Bump nixpkgs

### LINKED ISSUE #142: Increase upperbound on haskell tls package
Vaultenv is currently marked as `broken` in nixpkgs (24.05) because of the haskell dependency constraint : tls >=1.4 && <1.7

Would it be possible to release a new version that supports tls 1.8, the current latest version used in nixpkgs 24.05 ?

Thanks

alex-mckenna:
Bumped `nixpkgs`, and changed `connection` to `crypton-connection` since the former is deprecated. After this change, when I run `stack ls dependencies` in the nix shell I see that `tls` is bumped to `1.8.0`.

Closes #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ of `tutorial.sh`.
 ## Usage
 
 ```
-vaultenv 0.16.0 - run programs with secrets from HashiCorp Vault
+vaultenv 0.17.0 - run programs with secrets from HashiCorp Vault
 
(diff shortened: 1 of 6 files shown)


alex-mckenna:
Wait... we already had a PR which contains most of the fixes this will need that was just never released (#140). I'll fixup that instead.

PR #149, opened 2025-04-22
Add an alias for '--secrets-file'

RileyApeldoorn:
has this ever happened to you?

```
$ vaultenv -- echo "hello"
[ERROR] The option Secret file is required but not specified
$ vaultenv --secret-file foo.secrets -- echo "hello"
Invalid option `--secret-file'

Usage: vaultenv [--version] [--host HOST] [--port PORT] [--addr ADDR] 
                [--token TOKEN | --kubernetes-role ROLE | --github-token TOKEN] 
                [--secrets-file FILENAME] [CMD] [ARGS...] 
                [--no-connect-tls | --connect-tls] 
                [--no-validate-certs | --validate-certs] 
                [--no-inherit-env | --inherit-env] 
                [--inherit-env-blacklist COMMA_SEPARATED_NAMES] 
                [--retry-base-delay-milliseconds MILLISECONDS] 
                [--retry-attempts NUM] [--log-level error | info] [--use-path] 
                [--max-concurrent-requests NUM] 
                [--duplicate-variable-behavior error | keep | overwrite]
```

i also took the liberty of improving the "unspecified value" error message a little to mention the part of the USAGE that's missing directly

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,16 +101,17 @@ of `tutorial.sh`.
 ```
 vaultenv 0.18.0 - run programs with secrets from HashiCorp Vault
 
-Usage: vaultenv [--version] [--host HOST] [--port PORT] [--addr ADDR]
-                [--token TOKEN | --github-token TOKEN | --kubernetes-role ROLE]
-                [--secrets-file FILENAME] [CMD] [ARGS...]
-                [--no-connect-tls | --connect-tls]
-                [--no-validate-certs | --validate-certs]
-                [--no-inherit-env | --inherit-env]
-                [--inherit-env-blacklist COMMA_SEPARATED_NAMES]
-                [--retry-base-delay-milliseconds MILLISECONDS]
-                [--retry-attempts NUM] [--log-level error | info] [--use-path]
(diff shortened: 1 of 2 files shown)


Riscky (approved):
I think just improving the error message would do the trick, but I'm happy either way :)

FYI: you can specify the location of the secrets file in a `.env` file like this:
```
VAULTENV_SECRETS_FILE="foo/bar.secrets"
```
This prevents having to type out the location of the secrets file when running commands from the root of a project (or rather, the folder containing the .env file)

RileyApeldoorn:
sweet, thx for the tip! ^^

@OpsBotPrime merge

OpsBotPrime:
Rebased as eae5b8fa6b72c14134253062a13044f4f39497f4, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/6b57b908-b60a-43e0-b21e-c14e7054a30d?pipeline_id=6f91af6d-85a8-43a7-9c58-3d69267534ce) started.

PR #151, opened 2025-05-23
Make auth backend name configurable

koikonom:
Allows users to specify the name of their authentication backend if it's different than the default one.


README.md
```diff
@@ -125,6 +126,11 @@ Available options:
                            https://, the ip-address or DNS name, followed by the
                            port, separated with a ':'. Cannot be combined with
                            either VAULT_PORT or VAULT_HOST
+  --auth-backend AUTH_BACKEND
+                           Name of Vault authentication backend.Defaults to
+                           'kubernetes' or 'github' depending on the type of
+                           thechosen authentication method. Also configurable
```
ig11987, line 132: nit: `the chosen`

src/Config.hs
```diff
@@ -119,6 +120,7 @@ defaultOptions = Options
   { oVaultHost      = Just "localhost"
   , oVaultPort      = Just 8200
   , oVaultAddr      = parseURI "https://localhost:8200"
+  , oVaultAuthBackend = Just ""
```
Riscky, line 123: I think this should be `Nothing`, otherwise we would be sending requests to `/v1/auth//login` instead of `/v1/auth/github/login` and `/v1/auth/kubernetes/login` by default.

Riscky (changes requested):
Thanks for the PR!
We need to change the default, otherwise this LGTM

src/Config.hs
```diff
@@ -146,6 +148,7 @@ castOptions opts = Options
   { oVaultHost      = oVaultHost opts
   , oVaultPort      = oVaultPort opts
   , oVaultAddr      = oVaultAddr opts
+  , oVaultAuthBackend = oVaultAuthBackend opts
```
ruuda, line 151: We should also print the value in the `Show` instance below; it helps to clarify which value was selected in the end when verbose logs are enabled.

app/Main.hs
```diff
@@ -486,10 +486,12 @@ readKubernetesJwt =
 -- | Authenticate using GitHub auth, see https://www.vaultproject.io/docs/auth/github.
 requestGitHubVaultToken :: Context -> Text -> IO (Either VaultError ClientToken)
 requestGitHubVaultToken context ghtoken = let
+  backendName = if null oBackendName then "github" else oBackendName
+        where oBackendName = getOptionsValue oVaultAuthBackend (cCliOptions context)
```
ruuda, line 490:
```suggestion
  backendName = case oVaultAuthBackend (cCliOptions context) of
    Just name -> name
    Nothing -> "github"
```

If we change the default to value to `Nothing` as @Riscky suggested, to properly signal "not provided", instead of abusing empty string as a sentinel value, then here we can pick the backend name like this. [`getOptionsValue` would crash on the `Nothing`](https://github.com/channable/vaultenv/blob/eae5b8fa6b72c14134253062a13044f4f39497f4/src/Config.hs#L198-L200). Also below for the Kubernetes case.

It can also be written more concisely as 
```haskell
  backendName = fromMaybe "github" $ oVaultAuthBackend $ cCliOptions context
```
Personally I find the explicit pattern matching clearer.

app/Main.hs
```diff
@@ -486,10 +486,12 @@ readKubernetesJwt =
...
   bodyJson = Aeson.Object $ KM.singleton "token" (Aeson.String ghtoken)
   request = setRequestBodyJSON bodyJson
     $ setRequestMethod "POST"
-    $ unauthenticatedVaultRequest context "/v1/auth/github/login"
+    $ unauthenticatedVaultRequest context ("/v1/auth/" ++ backendName ++ "/login")
```
ruuda, line 494: (FYI, if we have an expression in parens as the final argument, then we can also use [the function application operator, `$`](https://www.stackage.org/haddock/lts-23.24/base-4.19.2.0/Prelude.html#v:-36-).

```suggestion
    $ unauthenticatedVaultRequest context
    $ "/v1/auth/" ++ backendName ++ "/login"
```

This is usually done to avoid closing parens piling up on the right-hand side, and it’s in a sense similar to method chaining in e.g. Rust. Both ways to write it are equivalent, both are fine. In this case, since there is already a chain of `$`s, maybe you find it prettier, it’s a matter of taste.

The full expression is equivalent to
```haskell
request = setRequestBodyJSON bodyJson
  (setRequestMethod "POST"
    (unauthenticatedVaultRequest context
      ("/v1/auth/" ++ backendName ++ "/login")))
```
which maybe makes it a bit clearer why people use `$` instead for deeply nested calls.)

README.md
ruuda: These missing spaces stem from the missing spaces in the string concatenation at https://github.com/channable/vaultenv/pull/151/files#diff-1a5d6b4436af44266b8d28b7df447ac4de73108f23a314c7e25512456552c915R611-R613.

src/Config.hs
```diff
@@ -597,6 +603,14 @@ optionsParser = Options
         <> help ("Vault address, the scheme, either http:// or https://, the ip-address or DNS name, " ++
             "followed by the port, separated with a ':'." ++
             " Cannot be combined with either VAULT_PORT or VAULT_HOST")
+    authBackend
+      = maybeStrOption
+      $ long "auth-backend"
+      <> metavar "AUTH_BACKEND"
+      <> value Nothing
+      <> help ("Name of Vault authentication backend."
+      ++ "Defaults to 'kubernetes' or 'github' depending on the type of the"
```
ruuda, line 612: The lack of spaces at the end of these strings cause the thing that @ig11987 pointed out in the readme.

koikonom:
For us to be able to print the authentication backed value we first need for it to be available when the config is built, so I moved the logic to that stage.

Riscky:
@OpsBotPrime merge

